PR #513, opened 2026-04-03
Extend classification

D00mch:
Take +1 messages from history for classification.
If the current message is too short, take even more messages from history.


D00mch:
@codex review

PR #514, opened 2026-04-04
Added gemma4 support

lovkikisa-gif:


composeApp/src/jvmMain/resources/scripts/build-llama-bridge.sh
```diff
@@ -53,9 +54,24 @@ resolve_llama_dir() {
   echo "$CACHED_LLAMA_DIR"
 }
 
+apply_local_llama_patch() {
```
D00mch, line 57: 1. Why is this needed? What is this patch?
2. Did you check whether this breaks Qwen?

native/llama-bridge/patches/llama.cpp-metal-bfloat-embed.patch
```diff
@@ -0,0 +1,62 @@
+diff --git a/ggml/src/ggml-metal/ggml-metal-device.m b/ggml/src/ggml-metal/ggml-metal-device.m
```
D00mch, line 1: Looks like a crutch. Let's fix the dylib directly (rebuild it) instead of applying patches. Or what I do not see here?

D00mch:
@codex review

composeApp/src/jvmMain/resources/scripts/build-llama-bridge.sh
lovkikisa-gif: This patch is needed because the bug is in the vendored `llama.cpp` Metal path, not in our Kotlin code, so simply rebuilding the dylib from the same sources would reproduce the same crash. 
It fixes the runtime Metal compilation settings for bf16 kernels and adds a guard so a missing pipeline fails cleanly instead of segfaulting. Long term I’d prefer to drop it by moving to an upstream revision with the fix, but for now this is the smallest reliable way to keep rebuilds reproducible.

PR #519, opened 2026-04-12
Local embeddings

lovkikisa-gif:


D00mch:
@codex review

composeApp/src/jvmMain/kotlin/ru/souz/ui/common/LocalModelUiSupport.kt
```diff
@@ -0,0 +1,96 @@
+package ru.souz.ui.common
```
D00mch: Let’s not populate namespace with the top level function names.
It’s not clear what this support file is for.
Let’s decide what abstraction this is and extract it into a class. Good for testing and extension in the future.

PR #539, opened 2026-05-03
Implemented runtime-aware sandbox selection end to end for the runtime/backend path.

D00mch:
Local mode remains the default. SOUZ_SANDBOX_MODE=local and a missing env var both resolve to local. SOUZ_SANDBOX_MODE=docker now selects Docker explicitly and fails fast if Docker or the required image is unavailable. Backend sandbox resolution is now per userId, and singleton tools resolve the active sandbox at invocation time from ToolInvocationMeta instead of capturing one sandbox instance.


D00mch:
@codex review

PR #547, opened 2026-05-16
Added image tools

lovkikisa-gif:


native/src/main/kotlin/ru/souz/llms/local/LocalModels.kt
```diff
@@ -51,10 +53,23 @@ data class LocalModelProfile(
     val promptFamily: LocalPromptFamily,
     val samplingDefaults: LocalSamplingDefaults,
     val useNativeGrammar: Boolean = false,
+    val visionProjectorCandidates: List<String> = emptyList(),
     override val licenseRequirements: LocalLicenseRequirements,
     override val defaultGpuLayers: Int = 99,
 ) : LocalDownloadableProfile
 
+data class LocalAuxiliaryDownloadProfile(
```
D00mch, line 71: Let's add docs to this class

D00mch, line 71: Let's add docs to this class

runtime/src/main/kotlin/ru/souz/llms/runtime/ChatResponseSupport.kt
```diff
@@ -0,0 +1,12 @@
+package ru.souz.llms.runtime
+
+import ru.souz.llms.LLMResponse
+
+internal fun LLMResponse.Chat.requireAssistantText(prefix: String): String = when (this) {
```
D00mch, line 11: need docs why this extension function exists

runtime/src/main/kotlin/ru/souz/llms/openai/OpenAIImageGenerationGateway.kt
```diff
@@ -0,0 +1,114 @@
...
+    private val apiKey: String
+        get() = settingsProvider.openaiKey
+            ?: System.getenv("OPENAI_API_KEY")
+            ?: System.getProperty("OPENAI_API_KEY")
+            ?: throw IllegalStateException("OPENAI_API_KEY is not set")
+
+    private val defaultImageModel: String
+        get() = System.getenv("OPENAI_IMAGE_MODEL")
+            ?: System.getProperty("OPENAI_IMAGE_MODEL")
+            ?: DEFAULT_IMAGE_MODEL
```
D00mch: No need to add additional sorce of data. We already store the selected model, this is one more source

runtime/src/main/kotlin/ru/souz/tool/RuntimeToolsModule.kt
```diff
@@ -160,6 +168,14 @@ class RuntimeToolsFactory(
             toolFindFolders.toGiga(),
         )
 
+        ToolCategory.IMAGE -> listOf(
+            toolViewImage.toGiga(),
+        )
+
+        ToolCategory.IMAGE_GENERATION -> listOf(
+            toolGenerateImage.toGiga(),
+        )
```
D00mch, line 177: Maybe one IMAGE category would be enough?

runtime/src/main/kotlin/ru/souz/llms/runtime/ImageCapabilities.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.nio.file.Path
+
+data class VisionInput(
+    val imagePath: Path,
+    val mimeType: String,
+    val sizeBytes: Long,
+    val question: String,
+)
+
+fun interface VisionGateway {
```
D00mch, line 24: Let's add docs on this. Why we need this abstraction.

native/AGENTS.md
```diff
@@ -38,3 +39,4 @@ Notes:
 - The rebuild script also reapplies `native/llama-bridge/patches/llama.cpp-metal-bfloat-embed.patch` before compiling, because the local `third_party/llama.cpp` checkout is gitignored and the Gemma 4 update needs an extra Metal runtime-compile fix on top of the pinned upstream ref.
 - On macOS the bridge now disables ggml Metal residency sets by default (`GGML_METAL_NO_RESIDENCY=1`) to avoid shutdown aborts; set `SOUZ_LLAMA_METAL_RESIDENCY=1` only if you need to opt back in for debugging.
 - Local runtime debug logs now include the effective generation variants (`context_size`, `max_tokens`, grammar flag) and native decode failures now return detailed batch/context parameters, so Gemma context-window issues can be diagnosed from app logs without attaching a debugger.
+- Multimodal generation now recalculates completion budget after `mtmd_tokenize`, using the real prompt/image token count instead of assuming media prompts consume zero context.
```
D00mch: `now` is useless in docs, it only makes sense during the time working on the task.

runtime/src/main/kotlin/ru/souz/tool/RuntimeToolsModule.kt
D00mch: Not sure about that.

D00mch:
ChatGPT review, the frist one seems to be critical:

Reviewed PR #547 (“Added image tools”) as a static PR review against the diff and the module AGENTS docs. I did not run Gradle/tests locally.

## Recommendation

**Hold merge until at least Findings 1–2 are addressed.** The feature direction looks good, but there are a couple of integration regressions around sandboxed file access and OpenAI attachment handling.

## Findings

### 1. **P1 — `ViewImage` bypasses sandbox/non-local file access**

`ToolViewImage` resolves the input through `FilesToolUtil.resolveSafeExistingFile`, but then falls back to direct JVM filesystem APIs and passes `Path.of(image.path)` into `VisionInput`; `OpenAIVisionGateway` then does `Files.readAllBytes(input.imagePath)`. This will work for local host paths, but it breaks the runtime sandbox abstraction, especially in Docker/non-local sandbox modes where the sandbox path is not necessarily readable from the host JVM. The runtime docs explicitly call out per-invocation sandbox resolution and avoiding direct host filesystem access when sandbox-aware abstractions exist. `FilesToolUtil` already has `withReadableLocalPath`, which is the right shape for local-or-temp bridging. ([[GitHub](https://github.com/D00mch/souz/blob/image-tools/runtime/src/main/kotlin/ru/souz/tool/files/ToolViewImage.kt)][1])

This is especially important because the PR exposes `IMAGE` as backend-safe, while backend/runtime tools may run under backend sandboxing rather than the desktop’s simple local path setup. ([[GitHub](https://github.com/D00mch/souz/pull/547/files)][2])

**Suggested fix:** make `ToolViewImage` use `filesToolUtil.withReadableLocalPath(image, meta) { localPath -> ... }` and pass that temporary/local path into the gateway only inside the block, or change `VisionInput` to carry bytes plus MIME type so provider gateways never reopen sandbox paths. Add a test with a fake/non-local sandbox where `localPathOrNull == null` to verify that `ViewImage` still succeeds.

---

### 2. **P1/P2 — OpenAI chat attachments now reject valid image URLs**

`OpenAIChatAPI.buildContent` now maps every attachment through `requireOpenAiImageDataUrl`, which only accepts strings starting with `data:image/` and throws for anything else. Before this PR, the attachment value was passed directly into `image_url.url`. The OpenAI Chat Completions reference says `image_url.url` can be either an image URL or base64 image data, and the image/vision guide also documents fully qualified URLs as a supported image-input form. ([[GitHub](https://raw.githubusercontent.com/D00mch/souz/refs/pull/547/head/runtime/src/main/kotlin/ru/souz/llms/openai/OpenAIChatAPI.kt)][3])

That means any existing caller that sends a public HTTPS image URL to OpenAI chat will now fail before reaching the API. The new test coverage appears to cover the new data URL path, but not the URL-preserving path. ([[GitHub](https://github.com/D00mch/souz/pull/547/files)][2])

**Suggested fix:** keep the data URL conversion inside `OpenAIVisionGateway` for local files, but let `OpenAIChatAPI` accept both `data:image/...` and `http(s)://...` image URLs. Add a regression test for `attachments = listOf("https://.../image.png")`.

---

### 3. **P2 — Screen-understanding classification exposes `StartScreenRecording` unintentionally**

The classification prompt now tells the classifier to choose both `DESKTOP` and `IMAGE` when a request needs a screenshot and then image understanding. The local classifier also scores “что видишь на экране / what is on the screen” for both `DESKTOP` and `IMAGE`. In the real desktop tool catalog, however, `DESKTOP` includes both `TakeScreenshot` and `StartScreenRecording`, while `IMAGE` includes `ViewImage`. So a normal “what do you see on my screen?” request will activate screen recording as an available tool even though the intended workflow is one screenshot plus image analysis. ([[GitHub](https://github.com/D00mch/souz/blob/image-tools/agent/src/main/kotlin/ru/souz/agent/nodes/NodesClassification.kt)][4])

The added test asserts only `TakeScreenshot` + `ViewImage`, but the test’s fake `DESKTOP` catalog contains only `TakeScreenshot`, so it does not catch the production behavior where `StartScreenRecording` is also present. ([[GitHub](https://github.com/D00mch/souz/pull/547/files)][2])

**Suggested fix:** split screenshot capture and screen recording into separate categories, or add a tool-level narrowing step for the `DESKTOP + IMAGE` workflow so “screen understanding” exposes only `TakeScreenshot` and `ViewImage`. Update the test to use the real desktop category mapping or include `StartScreenRecording` in the fake catalog to prevent this regression.

## Minor notes

The PR’s category split between `IMAGE` and `IMAGE_GENERATION` is a good direction, and the docs now describe the intended `TakeScreenshot -> ViewImage` chain clearly. ([[GitHub](https://github.com/D00mch/souz/blob/image-tools/AGENTS.md)][5])

I would also add one end-to-end-ish test for: “generated image path returned by `GenerateImage` can be passed into `ViewImage` under the same `ToolInvocationMeta`.” That would catch path-format and sandbox bridging problems early.

[1]: https://github.com/D00mch/souz/blob/image-tools/runtime/src/main/kotlin/ru/souz/tool/files/ToolViewImage.kt "souz/runtime/src/main/kotlin/ru/souz/tool/files/ToolViewImage.kt at image-tools · D00mch/souz · GitHub"
[2]: https://github.com/D00mch/souz/pull/547/files "Added image tools by lovkikisa-gif · Pull Request #547 · D00mch/souz · GitHub"
[3]: https://raw.githubusercontent.com/D00mch/souz/refs/pull/547/head/runtime/src/main/kotlin/ru/souz/llms/openai/OpenAIChatAPI.kt "raw.githubusercontent.com"
[4]: https://github.com/D00mch/souz/blob/image-tools/agent/src/main/kotlin/ru/souz/agent/nodes/NodesClassification.kt "souz/agent/src/main/kotlin/ru/souz/agent/nodes/NodesClassification.kt at image-tools · D00mch/souz · GitHub"
[5]: https://github.com/D00mch/souz/blob/image-tools/AGENTS.md "souz/AGENTS.md at image-tools · D00mch/souz · GitHub"

runtime/src/main/kotlin/ru/souz/tool/RuntimeToolsModule.kt
lovkikisa-gif: I’d keep them separate because IMAGE is for understanding an existing image, while IMAGE_GENERATION creates a new one, and merging them would expose GenerateImage in plain image-analysis flows where it is irrelevant.

PR #552, opened 2026-05-24
Memory new

lovkikisa-gif:


D00mch:
@codex review

D00mch:
@codex review

D00mch:
gpr pro review:

Reviewed PR #552. I found **4 issues worth fixing before merge**.

### 1. High — slot-key replacement can lose the existing memory on embedding/DB failure

`createCapturedFact` retires the currently active fact before creating and embedding the replacement: it finds the existing fact, calls `repo.retireFact(it.id)`, then calls `createFact(...)`. `createFact` inserts the new fact and only afterward calls `embedder.embedDocument(...)` / `repo.replaceEmbedding(...)`. If embedding fails, the old fact has already been retired and the replacement may be missing or unsearchable. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1]) ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1])

This same pattern affects edits: `updateFact` writes the fact first and then recalculates the embedding, so an embedding failure can leave changed text with a stale old vector. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1])

**Fix:** make fact mutation + embedding state atomic from the user-visible perspective. For replacement, create/embed the replacement first, then retire/supersede the old fact in the same DB transaction or with a compensating rollback path. For edits, either compute the embedding before committing the text update, or mark the embedding stale/missing and ensure retrieval re-embeds stale rows.

### 2. Medium — memory retrieval can block a normal chat turn by embedding every missing fact globally

Every agent execution builds the memory system prompt before the graph runs. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt)][2]) During `retrieveForPrompt`, the service calls `repo.getFactsWithoutEmbedding(embedder.model)` and embeds every returned fact before doing the actual scoped search. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1]) The SQLite implementation of `getFactsWithoutEmbedding` is not scoped; it selects all active facts missing that model’s embedding. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt)][3])

On first run after a model change, after import/migration, or after many failed captures, one user message can trigger a large synchronous embedding backfill before the assistant answers.

**Fix:** limit backfill to the requested scopes and a small batch size, move full reindexing to a background job, or let retrieval search already-indexed facts immediately and enqueue missing embeddings separately.

### 3. Medium — writer-provided scopes are trusted without validation

`MemoryCaptureInput` carries both `scopes` and `primaryScope`, but `MemoryCaptureService` stores `candidate.scope ?: input.primaryScope` and `isValidCandidate` only checks save flag, title/body/evidence, and confidence. It does not verify that `candidate.scope` is one of `input.scopes`. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt)][4]) ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryCaptureService.kt)][5])

Because candidate scopes come from the LLM writer output, this can store facts into invented or unintended scopes. Best case, those memories disappear from future retrieval because no runtime searches that scope; worst case, a prompt-injected turn can push a fact into a broader scope than intended.

**Fix:** reject candidate scopes not in `input.scopes`, or coerce unknown scopes to `input.primaryScope`. Add a test where the writer returns `MemoryScope("chat", "other-chat")` or a made-up scope and assert it is rejected.

### 4. Medium — empty embeddings are treated as successful embeddings

`LlmEmbeddingClient` returns `FloatArray(0)` when an embeddings response is `Ok` but has no first embedding. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1]) That empty vector can then be stored as an embedding row, and cosine similarity returns `0f` for zero-norm vectors. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt)][3]) ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt)][4]) Since retrieval filters to `score > 0f`, the fact will never be retrieved, and because an embedding row exists, the “missing embedding” backfill will not repair it. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt)][1])

**Fix:** treat empty embedding data as an error, do not persist zero-dimension embeddings, and consider validating dimension consistency before scoring. A regression test should cover `Embeddings.Ok(data = emptyList())` and assert that fact creation/update fails without writing an embedding row.

The UI metadata path looks okay: `ChatUseCase` passes `conversationId`, `requestId`, `userMessageId`, and `assistantMessageId` through `toolInvocationMetaOverride`, so I would not flag missing message IDs on the normal chat path. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/main/usecases/ChatUseCase.kt)][6])

[1]: https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt "souz/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryService.kt at memory-new · D00mch/souz · GitHub"
[2]: https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt "souz/agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt at memory-new · D00mch/souz · GitHub"
[3]: https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt "souz/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt at memory-new · D00mch/souz · GitHub"
[4]: https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt "souz/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt at memory-new · D00mch/souz · GitHub"
[5]: https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryCaptureService.kt "souz/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryCaptureService.kt at memory-new · D00mch/souz · GitHub"
[6]: https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/main/usecases/ChatUseCase.kt "souz/sharedUI/src/jvmMain/kotlin/ru/souz/ui/main/usecases/ChatUseCase.kt at memory-new · D00mch/souz · GitHub"

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -1,13 +1,29 @@
 package ru.souz.agent
 
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
```
D00mch: unused import

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -34,13 +50,121 @@ class AgentExecutor internal constructor(
...
+        val result = agentById(agentId).executeWithTrace(seed, onStep)
+        val normalizedResult = memoryAddedBlock
+            ?.let { result.copy(context = result.context.restoreSystemPrompt(baseSeed.systemPrompt)) }
+            ?: result
+
+        captureCompletedTurn(baseSeed, input, result.output)
+        return normalizedResult
+    }
+
+    private suspend fun buildMemorySystemPrompt(ctx: AgentContext<String>): ru.souz.memory.MemoryPromptAugmentationResult =
```
D00mch: ru.souz.memory.MemoryPromptAugmentationResult -> MemoryPromptAugmentationResult

desktopApp/src/main/kotlin/ru/souz/di/Dependencies.kt
```diff
@@ -117,13 +127,20 @@ val mainDiModule = DI.Module(DiTags.MODULE_MAIN) {
     bindSingleton<UiAudioRecorder> { instance<InMemoryAudioRecorder>() }
     bindSingleton<DesktopPermissionService> { MacDesktopPermissionService() }
 
-    // Native
+    // DB
+    bindSingleton<kotlinx.coroutines.CoroutineScope> { kotlinx.coroutines.CoroutineScope(kotlinx.coroutines.SupervisorJob() + kotlinx.coroutines.Dispatchers.Default) }
```
D00mch: Too long line.
Let's use imports instead of full paths.

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -34,13 +50,121 @@ class AgentExecutor internal constructor(
...
         onStep: GraphStepCallback?,
     ): AgentExecutionResult {
-        val seed = context.copy(
+        val runtimeEventSink = eventSink ?: context.runtimeEventSink
+        val baseSeed = context.copy(
             input = input,
-            runtimeEventSink = eventSink ?: context.runtimeEventSink,
+            runtimeEventSink = runtimeEventSink,
+        )
+        val memoryResult = buildMemorySystemPrompt(baseSeed)
```
D00mch: Let's rename `memoryResult` to `memoryPrompt`

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -34,13 +50,121 @@ class AgentExecutor internal constructor(
...
+        val augmentedSystemPrompt = memoryResult.augmentedSystemPrompt
+        val memoryAddedBlock = memoryAddedBlock(
+            baseSystemPrompt = baseSeed.systemPrompt,
+            augmentedSystemPrompt = augmentedSystemPrompt,
         )
-        return agentById(agentId).executeWithTrace(seed, onStep)
+        val seed = memoryAddedBlock
```
D00mch: Let's add `seed: AgentContext<String>` here to simplify reading.

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -34,13 +50,121 @@ class AgentExecutor internal constructor(
...
+        return addedBlock.takeIf { it.isNotBlank() }
     }
 
+    private fun AgentContext<String>.withSystemPrompt(prompt: String): AgentContext<String> =
+        copy(
+            systemPrompt = prompt,
+            history = history.replaceSystemPromptIfPresent(prompt),
+        )
+
+    private fun AgentContext<String>.restoreSystemPrompt(prompt: String): AgentContext<String> =
```
D00mch: Let's add docs why we need this.

agent/src/main/kotlin/ru/souz/memory/ConversationMemoryRuntime.kt
```diff
@@ -0,0 +1,44 @@
+package ru.souz.memory
```
D00mch, line 1: Let's add docs to alls the classes in the file

sharedLogic/src/main/kotlin/ru/souz/memory/MemoryWriter.kt
```diff
@@ -0,0 +1,240 @@
...
+}
+
+class LlmMemoryWriter(
+    private val api: LLMChatAPI,
+    private val settingsProvider: SettingsProvider,
+) : MemoryWriter {
+    private val logger = LoggerFactory.getLogger(LlmMemoryWriter::class.java)
+
+    override suspend fun extractCandidates(input: MemoryCaptureInput): List<MemoryFactCandidate> {
+        explicitRememberCandidate(input)?.let { return listOf(it) }
```
D00mch: Let's update this way, better for reading to see early exit 

```kotlin
val explicitRememberCandidate = explicitRememberCandidate(input)
if (explicitRememberCandidate != null) return listOf(explicitRememberCandidate)
```

That's also corresponds to your style below in `parseCandidates`

sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt
```diff
@@ -0,0 +1,270 @@
...
+        dot += a[index] * b[index]
+        normA += a[index] * a[index]
+        normB += b[index] * b[index]
+    }
+
+    if (normA == 0f || normB == 0f) return 0f
+    return dot / (sqrt(normA) * sqrt(normB))
+}
+
+object MemorySanitizer {
```
D00mch, line 222: Let's add docs

agent/src/main/kotlin/ru/souz/agent/AgentExecutor.kt
```diff
@@ -1,13 +1,29 @@
 package ru.souz.agent
```
D00mch, line 1: (!)
Let's move the logic to populate context with fact from here into the graph

D00mch:
@codex review

D00mch:
pro gpt review

1. Medium — prompt retrieval still does unbounded synchronous embedding backfill

The previous global backfill problem is improved, but not fully fixed. A normal chat turn still synchronously calls memory retrieval while constructing additional context: retrieveMemoryBlock awaits memoryRuntime.retrieveMemory(...), which calls memoryService.retrieveForPrompt(...). The runtime scopes always include GLOBAL_SCOPE, and include the chat scope when a conversation ID is available.

Inside retrieveForPrompt, before pinned/semantic retrieval, the service calls repo.getFactsWithoutEmbedding(distinctScopes, embedder.model).forEach { ... embedder.embedDocument(...) ... }. The SQLite implementation is scoped now, but still has no LIMIT, so one user message can embed every missing fact in global plus the current chat scope before the assistant responds.

This can still block a normal answer after import, migration, model change, or a period where embedding failed.

Suggested fix: make prompt-time repair bounded, for example embed at most 10–20 missing facts per request, or enqueue a background reindex job. Retrieval should use already-indexed facts immediately and repair missing embeddings separately.

2. Medium — pinned facts can silently disappear from the memory prompt

retrieveForPrompt tries to include pinned facts, but it gets them by calling repo.listFacts(MemoryFactFilter(scope = it, limit = limit)), then filtering it.pinned, then sorting/taking.

listFacts orders by updated_at DESC and applies the limit in SQL. That means an older pinned memory is not returned at all once enough newer facts exist in the same scope. The UI supports pinning, so this creates a misleading state: a user can pin a fact and still have it omitted from the prompt because it fell outside the first recent page.

Suggested fix: query pinned facts directly without the normal prompt limit, or add a pinnedOnly filter and use it in retrieveForPrompt. Another acceptable approach is ordering by pinned DESC, updated_at DESC, but a direct pinned query is cleaner.

3. Medium — edited memories bypass the sanitizer

Create/capture paths sanitize memory text before persistence: manual creation redacts title/body, and captured facts redact title/body/evidence before storage. The sanitizer is intended to redact secrets and private local data before memory is persisted.

The edit path does not do the same. MemoryEditorInput.toPatch() only trims title/body, and MemoryFact.applyPatch(...) also only trims title/body before updateFact embeds and persists the updated fact.

So a user can edit a memory and save text containing an API key, email, token, or local path that the create/capture paths would have redacted. That text can later be rendered into the memory prompt because renderMemoryPrompt uses the stored fact title/body directly.

Suggested fix: apply MemorySanitizer.redact in updateFact or applyPatch before embedding and repo.updateFact(...). Add a regression test that creates a safe memory, edits it to include something like apiKey=... or an email/path, and asserts the stored fact and embedding input are redacted.

Also worth fixing — cancellation can be swallowed during backfill

retrieveForPrompt wraps the missing-embedding backfill in runCatching, including the per-fact embedding call. In Kotlin, that pattern is risky around suspend work because cancellation is represented as an exception. If the user aborts a generation during a large backfill, the code can treat cancellation like an ordinary embedding failure and continue iterating.

Suggested fix: replace those runCatching blocks with try/catch and rethrow CancellationException; only suppress/log non-cancellation failures.

D00mch:
I reviewed PR #552 and would **request changes before merge**.

Context: this is a large memory feature PR: 16 commits from `memory-new`, 53 changed files, and about +5,802/−128 lines. It adds the desktop memory core and memory management UI. ([[GitHub](https://github.com/D00mch/souz/pull/552/files)][1]) The visible GitHub Actions status is also red: the JVM test job failed with exit code 1, and the page requires sign-in for the full logs, so I could only verify the public failure summary. ([[GitHub](https://github.com/D00mch/souz/actions/runs/26686112890/job/78654533860?pr=552)][2])

## Findings

### P1 — Real user prompts containing `<context>` can be deleted before the LLM call

`GraphBasedAgent` runs `inputToHistory -> ... -> contextEnrich -> LLM`, so the user’s current prompt is already in `ctx.history` before `nodeAppendAdditionalData` runs. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/GraphBasedAgent.kt)][3]) `inputToHistory` appends the actual user message to history, but `nodeAppendAdditionalData` then removes **every** user-role message whose content contains `"<context>"`. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/agent/nodes/NodesCommon.kt)][4]) Since `loadAdditionalData()` always appends current date/time for nonblank user text, a synthetic context message is inserted on essentially every turn, and `toGigaRequest` sends the modified history to the LLM. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/agent/nodes/NodesCommon.kt)][4])

A user asking something like `Explain this XML: <context>...</context>` would have their real prompt filtered out. In a first-turn case, this can also insert the synthetic context before the system message because only the system message remains and `newHistory.add(newHistory.lastIndex, it)` uses index `0`.

Suggested fix: do not identify generated context messages with a loose substring. Use a strict sentinel/prefix, metadata, or a dedicated wrapper helper, for example only remove messages that start with the exact generated context prefix and end with `</context>`. Add a regression test where the user prompt legitimately contains `<context>`.

### P2 — Embedding dimension mismatches are silently scored instead of rejected or rebuilt

The SQLite schema stores `dimension` for embeddings, and `upsertEmbedding` writes `embedding.size`, but `searchFacts` selects by `embedding_model` only and ignores the stored dimension. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt)][5]) It then calls `cosineSimilarity(queryEmbedding, row.embedding)`, while `cosineSimilarity` uses `minOf(a.size, b.size)` and scores only the overlapping prefix. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt)][5]) `getFactsWithoutEmbedding` also treats any row with the matching model as present, so a stale row with the right model name but wrong dimension will not be backfilled. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt)][5])

This can produce arbitrary retrieval results if an embedding provider changes dimensions under the same alias, a local embedding model is swapped, or a row is corrupted.

Suggested fix: require equal dimensions before scoring. Either filter with `e.dimension = ?` using `queryEmbedding.size`, or treat mismatched rows as stale and rebuild them. I’d also make `cosineSimilarity` return `0f` or throw on unequal sizes, and add a repository test for mismatched dimensions.

### P2/P3 — `DELETED` status and Deleted filter are unreachable because delete hard-removes rows

The domain model includes `MemoryFactStatus.DELETED`, and the UI exposes a `DELETED` filter plus `ALL` including deleted facts. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt)][6]) The details UI also has special logic to hide delete actions only once a fact is already `DELETED`. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryFactDialogs.kt)][7]) But confirming delete calls `memoryService.deleteFact`, and the repository implementation hard-deletes embeddings, evidence, and the `memory_facts` row. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryViewModel.kt)][8])

Result: users can select “Deleted”, but deleted facts can never appear there; `MemoryFactStatus.DELETED` is effectively dead state.

Suggested fix: choose one behavior. If delete is meant to be recoverable/auditable, implement it as `updateStatus(factId, DELETED)` and keep evidence/source events. If delete is meant to be permanent, remove `DELETED` from the enum/UI filters and simplify the details action logic.

### P3 — Path redaction only covers macOS `/Users/...` paths

`MemorySanitizer` says it redacts private local data before persistence, but the file path regex only matches `/Users/...`. Linux paths such as `/home/alice/project/.env` and Windows paths such as `C:\Users\Alice\...` would still be saved in memory source events. ([[GitHub](https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt)][6])

Suggested fix: broaden the path redaction patterns for Linux home paths, Windows user paths, and probably `~/.local/state/...` style paths. Add unit tests for each OS-style path.

[1]: https://github.com/D00mch/souz/pull/552/files "Memory new by lovkikisa-gif · Pull Request #552 · D00mch/souz · GitHub"
[2]: https://github.com/D00mch/souz/actions/runs/26686112890/job/78654533860?pr=552 "Memory new · D00mch/souz@f95df88 · GitHub"
[3]: https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/GraphBasedAgent.kt "souz/agent/src/main/kotlin/ru/souz/GraphBasedAgent.kt at memory-new · D00mch/souz · GitHub"
[4]: https://github.com/D00mch/souz/blob/memory-new/agent/src/main/kotlin/ru/souz/agent/nodes/NodesCommon.kt "souz/agent/src/main/kotlin/ru/souz/agent/nodes/NodesCommon.kt at memory-new · D00mch/souz · GitHub"
[5]: https://github.com/D00mch/souz/blob/memory-new/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt "souz/desktopApp/src/main/kotlin/ru/souz/memory/SqliteMemoryRepository.kt at memory-new · D00mch/souz · GitHub"
[6]: https://github.com/D00mch/souz/blob/memory-new/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt "souz/sharedLogic/src/main/kotlin/ru/souz/memory/MemoryModels.kt at memory-new · D00mch/souz · GitHub"
[7]: https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryFactDialogs.kt "souz/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryFactDialogs.kt at memory-new · D00mch/souz · GitHub"
[8]: https://github.com/D00mch/souz/blob/memory-new/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryViewModel.kt "souz/sharedUI/src/jvmMain/kotlin/ru/souz/ui/memory/MemoryViewModel.kt at memory-new · D00mch/souz · GitHub"

PR #569, opened 2026-07-22
feat: implement typing indicator for Telegram bot interactions

jilees:
Добавил в канал Телеграм индикатор typing при ожидании ответа от агента

### DIFF
--- a/backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotPollingService.kt
+++ b/backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotPollingService.kt
@@ -289,7 +289,6 @@ class TelegramBotPollingService(
         }
 
         try {
-            sendChatActionSafely(binding.id, token, message.chat.id)
             val result = coroutineScope {
                 val typingJob = launch { repeatTypingIndicator(binding.id, token, message.chat.id) }
                 try {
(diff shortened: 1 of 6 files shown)


D00mch:
@codex review

backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotPollingService.kt
```diff
@@ -288,13 +289,21 @@ class TelegramBotPollingService(
...
 
         try {
-            val result = turnExecutor.execute(
-                userId = binding.userId,
-                chatId = binding.chatId,
-                content = text,
-                clientMessageId = "telegram:${binding.id}:${update.updateId}",
-                requestOverrides = UserSettingsOverrides(streamingMessages = false),
-            )
+            sendChatActionSafely(binding.id, token, message.chat.id)
```
D00mch: **[P2] Keep the best-effort typing request off the turn's critical path**

This call is still awaited before `turnExecutor.execute`, so a slow Telegram `sendChatAction` delays agent processing by the full 35-second request timeout. There is a second delay path at `typingJob.cancelAndJoin()`: the backend implementation uses blocking `HttpClient.send`, which coroutine cancellation does not interrupt, so an in-flight periodic action can postpone the final reply as well. Start the indicator concurrently with the turn and use cancellable HTTP execution so this optional UI signal cannot delay execution or delivery. The desktop `withTypingIndicator` path also awaits its initial action before invoking the agent.

D00mch (commented):
Re-reviewed the latest head. The lease-fencing fix is correct and covered by a regression test; focused backend/desktop tests and CI pass. One P2 remains below.

backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotPollingService.kt
jilees: Тоже учел и обновил

PR #581, opened 2026-07-26
Restrict skills graph to core tools

D00mch:
## Summary

- Normalize every `SkillsGraphBasedAgent` input context to the three core tools before graph execution.
- Remove the `Install core tools` graph node and the obsolete `NodesCommon.installCoreTools` helper.
- Verify the first graph node receives only `GetSkills`, `GetKnowledge`, and `RunSkillCommand` in both advertised and executable tool collections.
- Update the agent invariants and skills-graph maintenance guidance.

## Why

The shared context can arrive populated with the complete tool catalog. Restricting tools inside the graph added an unnecessary setup node and allowed the initial graph state to contain unrelated functions. Applying the restriction at the agent boundary keeps the graph smaller and makes tool isolation true before any graph node runs, including for restored or manually supplied contexts.

## Impact

The skills-oriented agent exposes and executes only its three core tools. The classic graph and shared context factory retain their existing behavior.

## Validation

- `./gradlew :agent:test`
- `./gradlew :backend:test --tests 'ru.souz.backend.agent.runtime.BackendConversationRuntimeSettingsTest.runtime resolves skills graph with only its core tools'`
- `git diff --check`

### DIFF
--- a/agent/AGENTS.md
+++ b/agent/AGENTS.md
@@ -11,7 +11,7 @@ Before changing this module, read its [pain-point index](docs/pain-points.md) an
 ## Invariants
 
 - `AgentFacade` is a stateful, single-active-execution entry point. Starting a turn or changing its agent or context cancels the current execution. Model, prompt, temperature, and context-size setters currently update state in place, so callers must not use them concurrently with execution unless that lifecycle is changed and tested explicitly.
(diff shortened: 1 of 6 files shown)


D00mch:
@codex review

PR #593, opened 2026-07-30
Added search memory tool

lovkikisa-gif:


D00mch:
@codex review

agent/src/main/kotlin/ru/souz/agent/AgentCoreTools.kt
```diff
@@ -0,0 +1,37 @@
...
+    val runtimeCommandTool: LLMToolSetup,
+) {
+    val classicGraphTools: List<LLMToolSetup> = listOf(
+        getSkillByNameTool,
+        getKnowledgeTool,
+        searchKnowledgeTool,
+        searchMemoryTool,
+        runtimeCommandTool,
+    )
+    val skillsGraphTools: List<LLMToolSetup> = listOf(
```
D00mch: I don't like that we coupled two graph implementations with this data class. Maybe it would be better to pass this tools to the related graphs from dependency injection I don't have separate classes that tied them up.

agent/src/main/kotlin/ru/souz/memory/ConversationMemoryRuntime.kt
```diff
@@ -95,6 +95,32 @@ data class MemoryRetrievalResult(
     val trace: MemoryRetrievalTrace = MemoryRetrievalTrace(),
 )
 
+data class MemorySearchRequest(
```
D00mch: Can we avoid creating separate public data classes? The results that should be visible to the llm could be passed as just json object, maybe even as a string. These are able to add memory related facts to the context without these data classes. Why cannot we do something as we did before?

PR #594, opened 2026-07-30
Avoid auto-opening Calendar and Notes

D00mch:
## What changed

- Stop refreshing the macOS Calendar list during settings screen refresh.
- Load calendars lazily only when the default-calendar dropdown is opened.
- Avoid reading Notes during background desktop indexing unless Notes is already running.
- Add ViewModel coverage that settings refresh does not call the calendar provider.

## Why

Opening Souz could trigger macOS Calendar and Notes indirectly through startup/settings refresh and the background desktop indexer. These OS apps should only be touched when the user explicitly asks for calendar/notes actions or interacts with calendar selection.

## Impact

Souz startup is quieter and no longer launches Calendar or Notes as a side effect. Explicit Calendar and Notes tools keep working.

## Validation

- `./gradlew :sharedUI:jvmTest :desktopApp:test`
- `git diff --check`

### DIFF
--- a/desktopApp/src/main/kotlin/ru/souz/db/DesktopDataExtractor.kt
+++ b/desktopApp/src/main/kotlin/ru/souz/db/DesktopDataExtractor.kt
@@ -116,13 +116,19 @@ class DesktopDataExtractor(
     }
 
     fun notes(): List<StorredData> = runCatching {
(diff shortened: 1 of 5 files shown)


D00mch:
@codex review

PR #615, opened 2026-08-05
OAuth skill service for obtaining, storing, refreshing, and using OAuth tokens

jilees:
…ic core)

Lets skills call third-party APIs that require OAuth (Yandex to start, others via the same standard authorization-code flow later) without a raw access token ever reaching a skill's sandboxed process or the LLM's context.

- New :skill-oauth-api / :skill-oauth-impl modules: provider-agnostic SkillOAuthApi interface, AuthorizationCodeOAuthClient implementing the standard RFC 6749 authorization-code grant, Postgres-backed credential and pending-state repositories, AES-GCM token encryption at rest.
- ConnectOAuthProvider / CheckOAuthStatus / SafeApiCall tools bound through the shared skill manifest/approval-gate pipeline; a skill only ever declares oauthProvider/oauthScopes, never sees the token itself. Required scopes are enforced per-skill against the shared (userId, provider) credential, and outbound calls are restricted to each provider's declared allowedApiHosts — both defend against a hijacked model turn (e.g. via indirect prompt injection) misusing a broader or different provider grant.
- Provider registration is data-driven: OAuthProviderCatalog (oauth-providers.json, endpoints + allowed hosts only) plus per-provider <NAME>_OAUTH_CLIENT_ID/_CLIENT_SECRET/_REDIRECT_URI env vars for secrets; adding a standard-flow provider needs no Kotlin change.
- skill-oauth-impl owns its own Flyway migration location and schema history table, independent of the host backend's, so their version sequences can never collide. Baselines against the host's already- populated shared schema (baselineOnMigrate + baselineVersion "0") since the host's own migration always runs first against the same database.
- HTTP clients have explicit timeouts, correct Content-Type on outbound API calls, and are closed on backend shutdown.


backend/src/main/kotlin/ru/souz/backend/app/BackendAppConfig.kt
```diff
@@ -5,6 +5,17 @@ import ru.souz.backend.common.BackendConfigurationException
 import ru.souz.backend.config.BackendConfigSource
 import ru.souz.backend.config.BackendFeatureFlags
 import ru.souz.backend.config.SystemBackendConfigSource
+import ru.souz.skilloauth.impl.OAuthProviderCatalog
+
+/** Credentials for one [OAuthProviderCatalog] entry — read from `<NAME>_OAUTH_CLIENT_ID` /
```
D00mch: The comments got misaligned. The `/**` should be on its own line.

skill-oauth-api/src/main/kotlin/ru/souz/skilloauth/SkillOAuthApi.kt
```diff
@@ -0,0 +1,66 @@
...
+    val missingScopes: List<String> = emptyList(),
+)
+
+data class AuthorizationUrl(
+    val url: String,
+)
+
+data class ApiCallRequest(
+    val method: String,
+    val path: String,
```
D00mch: This seems to be "url", noth "path", right?

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    /**
+     * Handles the provider redirect: exchanges `code` for tokens and stores them.
+     * Intentionally not part of [SkillOAuthApi] — called only by the route installer in
+     * [installSkillOAuthRoutes], never by tool/agent code, since the callback is triggered by the
+     * provider's redirect rather than requested by any caller of this API.
+     */
+    internal suspend fun handleCallback(code: String, state: String): CallbackResult {
```
D00mch, line 266: If two skills from the same provider run one after another, there will be two callbacks here, and the last one will overwrite the first. Maybe we should merge the results here.

sharedLogic/src/commonJvmMain/kotlin/ru/souz/tool/skills/ToolSafeApiCall.kt
```diff
@@ -0,0 +1,91 @@
...
+ * the provider layer (`SkillOAuthApiImpl.requireAllowedApiUrl`) enforces HTTPS + a per-provider
+ * host allowlist before attaching the token — this tool must not be trusted to police that itself.
+ */
+class ToolSafeApiCall(
+    private val skillRegistryRepository: SkillRegistryRepository,
+    private val skillOAuthApi: SkillOAuthApi?,
+    private val approvalGate: SkillApprovalGate? = null,
+) : ToolSetup<ToolSafeApiCall.Input> {
+    data class Input(
+        @InputParamDescription("Activated Skill ID that declares an oauthProvider in its manifest.")
```
D00mch: ChatGPT reviewed this and is strongly suggesting changing the description to something like: "Skill ID whose approved manifest declares the OAuth provider and scopes for this operation", because the skill may not be activated.

I don't see much of a problem with it, but I'll dig into it later.

skill-oauth-api/src/main/kotlin/ru/souz/skilloauth/SkillOAuthApi.kt
D00mch: ChatGPT proposed to add headers as well and enum for the method. I don't insiist on enum, but headers may be a good idea:
```kotlin
enum class ApiMethod {
    GET,
    POST,
    PUT,
    PATCH,
    DELETE,
}

data class ApiCallRequest(
    val method: ApiMethod,
    val url: String,
    val headers: Map<String, String> = emptyMap(),
    val body: String? = null,
)

data class ApiCallResponse(
    val statusCode: Int,
    val headers: Map<String, List<String>> = emptyMap(),
    val body: String? = null,
)
```

D00mch:
Maybe rename the title as well, everything else is in English so far.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
jilees: Merging in the callback wouldn't be enough — the already-issued token from flow #1 isn't retroactively valid for scopes only requested in flow #2. Fixed it upstream instead: startAuthorization now supersedes any pending flow for the same (userId, provider) and folds its scopes into the new one, so there's only ever one live flow and the token that comes back always covers everything requested. Added a regression test + index.

D00mch: Agree, that's better

D00mch:
May I ask to to update PRs with new commits (no rebase)? We merge with squash anyway, but locally it's easier to review new changes when you add new commits.

skill-oauth-impl/src/test/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImplTest.kt
```diff
@@ -0,0 +1,300 @@
...
+        )
+
+        assertTrue(result.url.startsWith("https://oauth.yandex.ru/authorize?"))
+        assertTrue(result.url.contains("client_id=client-1"))
+        assertTrue(result.url.contains("scope=login%3Ainfo"))
+    }
+
+    @Test
+    fun `startAuthorization supersedes an existing pending state for the same user and provider`() = runTest {
+        // Regression test: without superseding, two overlapping ConnectOAuthProvider calls for the
```
D00mch, line 211: It seems that these connections aren't actually overlapping, but rather running synchronously.

D00mch: So this is possible and isn't tested:

1. Callback A removes pending state A.
2. Callback A starts the code → token exchange, but the credential has not been saved yet.
3. Skill B calls `status`:
   credential A is still missing.
4. `startAuthorization` for B runs:
   pending state A is already gone.
5. A new flow is created for B only.
6. Callback A saves token A.
7. Callback B saves token B and overwrites token A.

D00mch:
@codex review

skill-oauth-impl/src/main/resources/db/migration-skill-oauth/V1__skill_oauth.sql
```diff
@@ -0,0 +1,23 @@
...
+    state text primary key,
+    user_id text not null,
+    skill_id text not null,
+    provider text not null,
+    requested_scopes text not null default '',
+    expires_at timestamptz not null
+);
+
+create index skill_oauth_pending_states_expires_at_idx on skill_oauth_pending_states (expires_at);
+create index skill_oauth_pending_states_user_provider_idx on skill_oauth_pending_states (user_id, provider);
```
D00mch: Seems like we may have unique index here?

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+        runCatching { httpClient.close() }
+    }
+
+    override suspend fun status(userId: String, provider: String, requiredScopes: List<String>): OAuthStatus {
+        requireProviderClient(provider)
+        val credential = credentialRepository.find(userId, provider)
+        val granted = credential?.grantedScopes.orEmpty()
+        val missing = requiredScopes.filterNot { it in granted }
+        return OAuthStatus(
+            connected = credential != null && missing.isEmpty(),
```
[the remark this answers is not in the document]
D00mch: 1. The credential exists in the database.
2. The access token has expired.
3. The refresh token is missing, corrupted, or revoked.
4. `status()` still returns `connected=true`.
5. `ConnectOAuthProvider` responds with `"Already connected"`.
6. `SafeApiCall` keeps sending the invalid token.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+        skillId: String,
+        requiredScopes: List<String>,
+        request: ApiCallRequest,
+    ): ApiCallResponse {
+        val providerClient = requireProviderClient(provider)
+        val credential = credentialRepository.find(userId, provider)
+            ?: throw SkillOAuthException(
+                "Skill '$skillId' is not connected to '$provider'. Use ConnectOAuthProvider first."
+            )
+        val missingScopes = requiredScopes.filterNot { it in credential.grantedScopes }
```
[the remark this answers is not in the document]
D00mch: Imho, that's ok.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,237 @@
...
+        // and whichever lands last would upsert its own requestedScopes over the first one's,
+        // silently dropping scopes the first flow just had the user grant. Folding the old pending
+        // scopes into this new request keeps the link the user ends up completing (whichever one
+        // that is) always asking for the full set requested so far; the superseded link, if opened
+        // afterwards, fails cleanly as an invalid/expired state rather than corrupting the grant.
+        val supersededScopes = pendingStateRepository.consumeActiveForUserAndProvider(userId, provider, now)
+            .flatMap { it.requestedScopes }
+        val mergedScopes = (supersededScopes + scopes).distinct()
+        val state = generateState()
+        pendingStateRepository.create(
```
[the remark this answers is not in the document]
D00mch: I wrote the same inside the test related comment section.

backend/src/main/kotlin/ru/souz/backend/http/BackendHttpServer.kt
```diff
@@ -178,6 +179,7 @@ internal fun Application.configureBackendHttpServer(dependencies: BackendHttpDep
         }
 
         v1Routes(dependencies)
+        dependencies.skillOAuthApiImpl?.let { installSkillOAuthRoutes(it, BackendHttpRoutes.OAUTH_CALLBACK) }
```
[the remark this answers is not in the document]
D00mch: We can avoid doing that as we expect the future implementation to be in go

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthCredentialRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+                """
+                insert into skill_oauth_credentials(
+                    user_id, provider, access_token_encrypted, refresh_token_encrypted,
+                    granted_scopes, expires_at, created_at, updated_at
+                )
+                values (?, ?, ?, ?, ?, ?, ?, ?)
+                on conflict (user_id, provider) do update
+                set access_token_encrypted = excluded.access_token_encrypted,
+                    refresh_token_encrypted = excluded.refresh_token_encrypted,
+                    granted_scopes = ${mergedScopesColumnSql("skill_oauth_credentials", "granted_scopes")},
```
D00mch: **Keep the stored scopes coupled to the stored token.** The normal sequential path requests the union correctly, but the reviewed interleaving still breaks that invariant: callback A can consume its pending state and start exchanging while no credential exists; B then sees no credential and requests only scope B. Whichever callback writes last supplies one access token, while this SQL unions both callbacks' scope metadata. The row can therefore claim `A + B` while the stored token only carries A or B. Please keep in-flight requested scopes/versioning until completion, prevent a superseded callback from overwriting a newer token, and store `grantedScopes` only for the access token that is actually persisted.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+     * to *proactively* refresh — this one only asks whether recovery is possible at all, which is
+     * what [status] and [callAuthorizedApi] need: a credential that merely needs a refresh soon is
+     * still "connected"; one that has expired with no way to refresh is not, and must not be
+     * reported as connected just because a row for it exists (that would leave `ConnectOAuthProvider`
+     * permanently reporting "already connected" instead of ever issuing a fresh authorize URL).
+     */
+    private fun isCredentialUsable(credential: SkillOAuthCredential): Boolean {
+        val expiresAt = credential.expiresAt ?: return true
+        if (expiresAt.isAfter(clock.instant())) return true
+        return credential.refreshTokenEncrypted != null
```
D00mch, line 315: **A stored refresh token is not necessarily a usable refresh token.** A corrupted or revoked refresh token still makes `status()` return connected; when refresh later fails with `invalid_grant`, `ConnectOAuthProvider` continues returning "Already connected" and there is no recovery path. Please add an explicit forced-reconnect path and invalidate/mark the credential unusable on a confirmed refresh `invalid_grant` (rather than deleting it for an arbitrary resource API 401).

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,261 @@
...
+
+    private suspend fun ensureFreshAccessToken(
+        credential: SkillOAuthCredential,
+        providerClient: OAuthProviderClient,
+    ): String {
+        val expiresAt = credential.expiresAt
+        if (expiresAt == null || expiresAt.isAfter(clock.instant().plusSeconds(EXPIRY_SAFETY_MARGIN_SECONDS))) {
+            return crypto.decrypt(credential.accessTokenEncrypted)
+        }
+        val refreshTokenEncrypted = credential.refreshTokenEncrypted
```
D00mch: **Avoid the refresh-margin dead zone when no refresh token exists.** `status()` treats the credential as connected until the actual `expiresAt`, but `ensureFreshAccessToken()` enters this branch up to 60 seconds earlier and throws when `refreshTokenEncrypted` is null. During that window API calls fail while reconnect still says "Already connected". If no refresh token exists, either use the access token until its actual expiry or make `status()` use the same threshold.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthPendingStateRepository.kt
```diff
@@ -0,0 +1,63 @@
...
+            connection.prepareStatement(
+                """
+                insert into skill_oauth_pending_states(
+                    state, user_id, skill_id, provider, requested_scopes, expires_at
+                )
+                values (?, ?, ?, ?, ?, ?)
+                on conflict (user_id, provider) do update set
+                    state = excluded.state,
+                    skill_id = excluded.skill_id,
+                    requested_scopes = ${mergedScopesColumnSql("skill_oauth_pending_states", "requested_scopes")},
```
D00mch: **Do not merge scopes from an expired pending state.** An expired row remains until its callback is consumed, and this conflict update merges its scopes without checking `expires_at`. A later authorization can therefore request obsolete permissions. Make the upsert treat an expired existing row as absent—replace its scopes instead of unioning them—while retaining the single-statement atomicity.

D00mch (commented):
Follow-up review of the OAuth race and credential-recovery fix. The atomic pending-state upsert is a good improvement; the inline notes cover the remaining correctness gaps we discussed.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthCredentialRepository.kt
jilees: Agreed — removed the scope merge entirely. grantedScopes is now a plain replace, guarded by a monotonic per-(userId, provider) generation snapshotted at authorization-start (not at callback completion, which would degenerate to last-write-wins): a stale write can never clobber a fresher credential. The actual race is now closed upstream — a new SkillOAuthRequestedScopesRepository tracks the cumulative union of requested scopes independent of any pending state's lifecycle, so a second authorization still widens on top of an in-flight one even after its pending state is consumed. In the common case both flows converge on the same request before either completes, so nothing real gets lost.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
jilees: On a confirmed OAuth error from refresh (not a network failure — different exception type, not caught) we now clear the refresh token, so status() correctly flips to disconnected instead of retrying a dead token forever. For the reconnect path: callAuthorizedApi now returns a fresh authorize URL directly as part of its normal response (ApiCallReconnectRequired) whenever the credential turns out unusable, rather than throwing — so SafeApiCall can relay it in one step, no separate ConnectOAuthProvider call needed.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthPendingStateRepository.kt
jilees: This exact merge is gone — pending-state upsert no longer merges scopes at all (moved to the new SkillOAuthRequestedScopesRepository, see #1). That table has the same kind of concern though: it has no TTL of its own, so I added updated_at and treat a row untouched longer than the pending-link TTL as abandoned — its old scopes are dropped, not folded into a later, unrelated authorization.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthPendingStateRepository.kt
```diff
@@ -0,0 +1,66 @@
...
+                """
+                insert into skill_oauth_pending_states(
+                    state, user_id, skill_id, provider, requested_scopes, generation, expires_at
+                )
+                values (?, ?, ?, ?, ?, ?, ?)
+                on conflict (user_id, provider) do update set
+                    state = excluded.state,
+                    skill_id = excluded.skill_id,
+                    requested_scopes = excluded.requested_scopes,
+                    generation = excluded.generation,
```
D00mch, line 90: **[P1] Keep generation allocation and pending-state replacement atomic.** `mergeAndBump()` commits before this separate upsert. A can obtain generation 1 and pause; B can obtain generation 2 and store its broader pending state; then A resumes and this unconditional conflict update replaces B with generation 1. B’s newly returned URL is invalidated while A’s narrower state remains active. Please combine the generation/scope update and pending-state replacement in one transaction, or add equivalent generation rejection plus retry/return handling so an older start cannot supersede a newer one.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,342 @@
...
+     * supersede) a second link for the same turn.
+     */
+    private suspend fun reconnectRequired(
+        userId: String,
+        provider: String,
+        skillId: String,
+        requiredScopes: List<String>,
+        reason: String,
+    ): ApiCallReconnectRequired {
+        val authorization = startAuthorization(userId, provider, skillId, requiredScopes)
```
D00mch: **[P2] Preserve the credential’s existing grants when generating this reconnect URL.** This passes only the current skill’s `requiredScopes`. Once `skill_oauth_requested_scopes.updated_at` is older than the 10-minute window, `mergeAndBump()` drops its previous scopes, so a credential with scope A reconnecting for skill B requests only B. The callback then replaces the credential with a B-only token and disconnects skill A. Include `credential.grantedScopes + requiredScopes` when a credential exists, or otherwise give completed grants a lifecycle distinct from abandoned requests.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthApiImpl.kt
```diff
@@ -0,0 +1,342 @@
...
+        if (stillFresh) {
+            return crypto.decrypt(credential.accessTokenEncrypted)
+        }
+        val refreshToken = refreshTokenEncrypted
+            ?: throw SkillOAuthException(
+                "OAuth access token for '${credential.provider}' has expired and no refresh token is available."
+            )
+        val refreshed = try {
+            providerClient.refresh(crypto.decrypt(refreshToken))
+        } catch (e: SkillOAuthException) {
```
D00mch: **[P2] Invalidate the refresh token only for a structured `invalid_grant`.** `AuthorizationCodeOAuthClient.toTokenResult()` converts every parsed OAuth error without an `access_token` into the same `SkillOAuthException`, so this catch also clears the token for `invalid_client`, `invalid_scope`, and provider-specific temporary errors. Those are not evidence that the user’s refresh token is dead, and a new consent flow may not fix them. Preserve the OAuth error code/status and clear the stored refresh token only for errors that confirm token revocation/invalidity.

D00mch (commented):
Follow-up review of the OAuth concurrency and reconnect fixes. The previous scope/token coupling and refresh-margin issues are improved, but these three cases remain reproducible.

P.S. 
AI Generated.

D00mch:
@codex review

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthCredentialRepository.kt
```diff
@@ -0,0 +1,83 @@
...
+                )
+                values (?, ?, ?, ?, ?, ?, ?, ?, ?)
+                on conflict (user_id, provider) do update
+                set access_token_encrypted = excluded.access_token_encrypted,
+                    refresh_token_encrypted = excluded.refresh_token_encrypted,
+                    granted_scopes = excluded.granted_scopes,
+                    expires_at = excluded.expires_at,
+                    generation = excluded.generation,
+                    updated_at = excluded.updated_at
+                where excluded.generation >= skill_oauth_credentials.generation
```
D00mch: [P1] Prevent equal-generation refresh writes from clobbering one another. Authorization generations order callbacks, but concurrent refreshes intentionally reuse the same generation and this `>=` accepts both. If refresh A succeeds and stores a rotated access/refresh token, while refresh B (which used the same old refresh token) then gets `invalid_grant`, B's `credential.copy(refreshTokenEncrypted = null)` is also accepted at the same generation and overwrites A's fresh row with the original expired access token and no refresh token. Please use an optimistic token/version check for refresh updates and invalidation (for example, clear only if the stored encrypted refresh token/version still matches what this call read).

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,351 @@
...
+        requireAllowedApiUrl(providerClient, request.url)
+        val accessToken = try {
+            ensureFreshAccessToken(credential, providerClient)
+        } catch (e: SkillOAuthException) {
+            // Both of ensureFreshAccessToken's own failure modes (no refresh token left, or a
+            // refresh token the provider just confirmed is invalid) mean the same thing here:
+            // reconnecting is the only way forward. A transient network failure reaching the
+            // provider throws a *different* exception type and is deliberately not caught here —
+            // that's not something a fresh authorize link would fix.
+            return reconnectRequired(
```
D00mch, line 119: [P2] Do not turn non-`invalid_grant` refresh errors into reconnects. `ensureFreshAccessToken()` now preserves the token for `server_error`, `invalid_client`, etc., but wraps all provider errors back into `SkillOAuthException`; this catch handles every such exception by calling `reconnectRequired()`. A new consent flow cannot fix a provider outage or client-secret misconfiguration, and it also supersedes the current pending authorization unnecessarily. Preserve the structured error classification through this boundary and only produce `ApiCallReconnectRequired` for `invalid_grant`/locally unusable credentials; surface other errors as transient or configuration failures.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,351 @@
...
+                    header(name, value)
+                }
+            }
+            header(HttpHeaders.Authorization, "Bearer $accessToken")
+            apiRequest.body?.let {
+                header(HttpHeaders.ContentType, ContentType.Application.Json.toString())
+                setBody(it)
+            }
+        }
+        return ApiCallResponse(
```
D00mch, line 153: [P2] Keep an explicit forced-reconnect path for revoked access tokens. A resource API 401 is returned as a normal `ApiCallResponse`; because a non-expired credential still makes `status()` report connected, `ConnectOAuthProvider` then returns “Already connected” and the user cannot obtain a fresh authorization URL. Automatically invalidating every resource 401 would be unsafe, but the connect API/tool needs a deliberate force-reconnect option (or an equivalent user-triggered invalidation path).

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthPendingStateRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+                    resultSet.next()
+                    Triple(
+                        resultSet.getString("requested_scopes").fromScopesColumn(),
+                        resultSet.getLong("generation"),
+                        resultSet.instant("updated_at"),
+                    )
+                }
+            }
+
+            val baseScopes = if (updatedAt.isBefore(activeSince)) emptyList() else existingScopes
```
D00mch, line 58: [P2] Do not classify an authorization as abandoned solely from its start time. A callback can consume its state near the end of the 10-minute TTL and still be exchanging the code when a second authorization begins just after the cutoff. This branch then drops the first flow's scopes even though it is actively completing; the newer generation can store a narrower token while the first callback still reports success. Track or refresh the authorization lifecycle when callback processing begins (or explicitly retain in-flight scopes) before applying the abandoned-flow reset.

D00mch (commented):
Follow-up review of `2a1f0296`. The atomic begin-authorization, existing-scope preservation, and structured OAuth-error changes are improvements, and the affected Gradle suites pass. I found four remaining refresh/recovery lifecycle issues, detailed inline.

D00mch:
I see a good opportunity to simplify this, especially if the OAuth implementation is expected to move to a separate service.

The main simplification would be to stop modeling per-Skill isolation unless we actually intend to enforce it. The current architecture stores one grant per `(userId, provider)`, so `skillId` and `requiredScopes` do not create a real per-Skill security boundary. I would either introduce a real Skill-bound capability/grant or—preferably for the current requirements—make the contract honestly user/provider-scoped.

For the simpler model, the API could be reduced to something like:

```kotlin
interface SkillOAuthGateway {
    suspend fun ensureAuthorized(
        userId: String,
        provider: String,
        requiredScopes: Set<String>,
        force: Boolean = false,
    ): AuthorizationState

    suspend fun call(
        userId: String,
        provider: String,
        requiredScopes: Set<String>,
        request: ApiCallRequest,
    ): ApiCallOutcome
}

sealed interface AuthorizationState {
    data object Connected : AuthorizationState
    data class AuthorizationRequired(val url: String) : AuthorizationState
}
```

That would allow us to remove:

- the separate `status()` + `startAuthorization()` pair; check-and-start should be one service-owned operation;
- `skillId` from the OAuth service contract (unless it is retained only as audit metadata);
- `OAuthStatus.grantedScopes`, since scope union belongs inside the service;
- `CheckOAuthStatus`, because `ConnectOAuthProvider` can return `Connected`, while `SafeApiCall` already returns reconnect-required;
- duplicated scope-union logic in the tool and service layers;
- repeated `.distinct()` calls by representing scopes as `Set<String>`.

I would also simplify the integration:

- construct OAuth tools only when a non-null `SkillOAuthApi` exists; remove the duplicated `SkillOAuthApi?` + `hasSkillOAuthApi` state;
- expose them through one optional OAuth tool catalog/module rather than always creating unavailable portable tools;
- resolve `skillId -> effective Skill -> provider/scopes` once through the same compiled/file-backed precedence used by normal Skill discovery;
- keep `ToolCategory.OAUTH`, but defer classifier examples/regex until a classifier-based host actually exposes OAuth tools;
- isolate the temporary local backend wiring behind one `skillOAuthBackendModule(config)`, so replacing it with a microservice client affects one composition point;
- parse `oauthScopes` as real YAML, or reject unsupported syntax instead of silently treating it as an empty list.

I would keep `ApiCallOutcome.ReconnectRequired`, the rule that access tokens never cross the contract, and URL/host validation next to Authorization-header injection.

If true per-Skill isolation is required, the simpler contract above is insufficient: we would need a service-issued capability bound to an immutable Skill identity and an enforceable endpoint/method policy. The current design is in an awkward middle ground—most of the complexity of per-Skill authorization without an actual per-Skill boundary.

### Two concrete P2 issues related to this simplification

There are also two concrete defects behind these suggestions:

- `SkillBundleParser.kt`: `oauthScopes` is recognized only in one exact block-list shape. A valid inline list, or a comment/blank line before the first item, can silently become an empty scope list. The tool may then reuse a broader existing credential without enforcing the manifest's intended restriction. This should be parsed as YAML, or unsupported syntax should fail explicitly instead of becoming `emptyList()`.
- `BackendDiModule.kt`: the OAuth API and tools are enabled when the encryption key is configured even if there are no configured OAuth providers. The tools are then advertised to the agent, but every invocation fails with `Unsupported OAuth provider`. OAuth capability should be installed only when at least one valid provider is configured, rather than being derived from the encryption key or a merely non-null API instance.

skill-oauth-api/src/main/kotlin/ru/souz/skilloauth/SkillOAuthGateway.kt
```diff
@@ -0,0 +1,102 @@
...
+ * sandboxed process or the LLM's context. The inbound OAuth redirect callback is deliberately not
+ * part of this interface — it is triggered by the provider, not by any caller of this API — see
+ * the route-installer function in `:skill-oauth-impl` instead.
+ *
+ * All methods are `suspend` so that a local, in-process implementation can later be swapped for a
+ * thin HTTP/gRPC client talking to an extracted microservice without any change to callers.
+ */
+interface SkillOAuthGateway {
+    /**
+     * Idempotent check-and-start: returns [AuthorizationState.Connected] when the shared credential
```
D00mch, line 24: **[P2] `ensureAuthorized` invalidates an equivalent active authorization link.**

The contract describes the operation as idempotent, but every call made before credentials exist creates a new state and invalidates the previous URL. A retry or overlapping request can therefore invalidate a link already returned to the user.

While an unexpired, unconsumed authorization is pending, atomically reuse its state and URL when it covers the same effective scope union. Create a new link only when scopes widen, the pending request expires or is consumed, or `force=true`; reusing it should not extend its TTL.

agent/src/main/kotlin/ru/souz/agent/skills/bundle/SkillBundleParser.kt
```diff
@@ -46,11 +46,65 @@ object SkillBundleParser {
...
+        }
+
+        val values = mutableListOf<String>()
+        var lineIndex = keyLineIndex + 1
+        while (lineIndex < lines.size) {
+            val line = lines[lineIndex]
+            val trimmed = line.trim()
+            when {
+                trimmed.isEmpty() || trimmed.startsWith("#") -> lineIndex++
+                !line.startsWith("  ") -> return values
```
D00mch: **[P2] Valid `oauthScopes` YAML can still silently become an empty list.**

The block-list parser requires two literal leading spaces, so valid YAML such as `oauthScopes:\n - login:info` returns `emptyList()` rather than parsing or rejecting the value. A new authorization then omits the scope declared by the skill, preventing the subsequent API operation from obtaining its required grant.

Parse this field using a YAML parser, or reject every present list shape the custom parser cannot handle, and add coverage for one-space indentation and null/bare values.

D00mch (commented):
Two remaining P2 findings from the OAuth contracts and integration re-review.

D00mch:
Can you also mention the new modules in the `AGENTS.md`?

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,401 @@
...
+    private suspend fun startAuthorization(
+        userId: String,
+        provider: String,
+        scopes: Set<String>,
+        force: Boolean = false,
+    ): String {
+        val providerClient = requireProviderClient(provider)
+        val now = clock.instant()
+        if (!force) {
+            val active = pendingStateRepository.findActive(userId, provider, now)
```
D00mch, line 184: **[P2] Make pending-link reuse atomic with creation**

This fixes sequential retries, but `findActive` and `beginAuthorization` are still separate repository operations/transactions. Two genuinely concurrent calls for the same `(userId, provider)` can therefore interleave as follows:

1. Request A calls `findActive` and sees no pending state.
2. Request B calls `findActive` before A creates one and also sees no pending state.
3. A enters `beginAuthorization`, creates `state-A`, and returns `url-A`.
4. B enters `beginAuthorization`; the repository lock serializes the write, but B still unconditionally replaces `state-A` with `state-B` and returns `url-B`.
5. A caller that received `url-A` now gets an invalid/expired-state response from the callback.

The test named `two overlapping calls...` invokes the calls one after another, so it only verifies the sequential case and cannot expose this race. Please move the reuse/coverage decision into the same per-user/provider repository transaction that creates or widens the pending state. That operation should return either the existing covering state or the newly created state, from which this method can build the URL.

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/PostgresSkillOAuthPendingStateRepository.kt
```diff
@@ -0,0 +1,166 @@
...
+                statement.setLong(6, nextGeneration)
+                statement.setInstant(7, expiresAt)
+                statement.executeQuery().use { resultSet ->
+                    resultSet.next()
+                    resultSet.toPendingState()
+                }
+            }
+        }
+
+    override suspend fun consume(state: String, now: Instant): SkillOAuthPendingState? =
```
D00mch, line 109: **[P2] Use one lock order for authorization start and callback consumption**

`beginAuthorization` first locks the `skill_oauth_requested_scopes` row with `SELECT ... FOR UPDATE`, then upserts the matching `skill_oauth_pending_states` row. `consume` takes the reverse order: `DELETE ... RETURNING` locks/deletes the pending row first, then this transaction updates the requested-scopes row.

For the same `(userId, provider)`, a callback can race with a forced or scope-widening authorization like this:

1. T1 (`beginAuthorization`) locks the requested-scopes row R.
2. T2 (`consume`) deletes/locks pending row P, then blocks while updating R.
3. T1 reaches `INSERT ... ON CONFLICT DO UPDATE` for P and blocks on T2's uncommitted delete.
4. PostgreSQL detects the R→P / P→R cycle and aborts one transaction with SQLSTATE `40P01`.

`DataSource.write` does not retry deadlocks, so this can surface as a failed tool call or an OAuth callback failure after the user already completed consent. Please make both paths acquire locks in the same order—preferably the pair/requested-scopes lock before touching the pending row. A bounded retry for `40P01` can be additional protection, but it should not replace the consistent lock order.

agent/src/main/kotlin/ru/souz/agent/skills/bundle/SkillBundleParser.kt
```diff
@@ -46,11 +46,72 @@ object SkillBundleParser {
...
+            val trimmed = line.trim()
+            when {
+                trimmed.isEmpty() || trimmed.startsWith("#") -> lineIndex++
+                // Any positive indentation nests under the key — YAML doesn't mandate a specific
+                // width, and hardcoding one here (as this used to) meant a validly-indented single-
+                // space list like 'oauthScopes:\n - a' silently parsed as zero items instead of
+                // either accepting or rejecting it.
+                !line.startsWith(" ") && !line.startsWith("\t") -> return values
+                trimmed.startsWith("- ") -> {
+                    values += trimmed.removePrefix("-").trim().trim('"', '\'')
```
D00mch, line 89: **[P2] Parse YAML trailing comments instead of including them in scopes**

This accepts one-space indentation now, but valid YAML end-of-line comments are still interpreted as scope text. For example:

```yaml
oauthScopes:
  - login:info # basic profile access
```

produces the single stored string `login:info # basic profile access`. `buildAuthorizeUrl` later joins scopes with spaces, so the provider receives extra scope tokens such as `#`, `basic`, `profile`, and `access`, commonly resulting in `invalid_scope` or an authorization that does not match the manifest.

The equivalent flow form is also valid YAML but currently rejected because the raw value no longer ends with `]`:

```yaml
oauthScopes: [login:info] # basic profile access
```

Standalone comment-line tests do not cover either case. Please parse this sequence structurally with a YAML parser, or make the custom parser quote- and comment-aware, and add regression tests for trailing comments in both block and flow forms.

D00mch (commented):
I found three remaining P2 correctness issues: the pending-link reuse decision is not atomic with creation, the pending-state repository uses an inverted lock order that can deadlock a callback against a new authorization, and the hand-written YAML list parser mishandles valid trailing comments. Details and concrete interleavings are in the inline comments.

D00mch:
@codex review

D00mch:
@codex review

skill-oauth-impl/src/main/kotlin/ru/souz/skilloauth/impl/SkillOAuthGatewayImpl.kt
```diff
@@ -0,0 +1,397 @@
...
+            // determines the actual outgoing Content-Type from the request body's own metadata,
+            // not from a plain `header()` call, so setting it that way here wouldn't reliably work.
+            apiRequest.headers.forEach { (name, value) ->
+                if (!name.equals(HttpHeaders.Authorization, ignoreCase = true) &&
+                    !name.equals(HttpHeaders.ContentType, ignoreCase = true)
+                ) {
+                    header(name, value)
+                }
+            }
+            header(HttpHeaders.Authorization, "Bearer $accessToken")
```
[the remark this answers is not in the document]
jilees: In fact both providers works fine, but anyway will add schema to be sure

D00mch:
@codex review

D00mch:
Local codex:

Reviewed PR #615 at `8d4dc2f0`, focusing on contracts and existing code paths rather than `skill-oauth-impl`.

**Findings**

- [P2] [`[SkillBundleParser.kt#L114](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/agent/src/main/kotlin/ru/souz/agent/skills/bundle/SkillBundleParser.kt#L114)`](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/agent/src/main/kotlin/ru/souz/agent/skills/bundle/SkillBundleParser.kt#L114): inline `oauthScopes` still splits blindly on commas. Valid YAML like `oauthScopes: ["resource,read"]` becomes two scopes, `resource` and `read`. Scope tokens are provider-defined, so the manifest contract should preserve them exactly. A real YAML parser is the cleanest fix; otherwise the flow-list parser needs quote/escape awareness and a regression test.

- [P3] [`[BackendDiModuleTest.kt#L163-L174](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/backend/src/test/kotlin/ru/souz/backend/app/BackendDiModuleTest.kt#L163-L174)`](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/backend/src/test/kotlin/ru/souz/backend/app/BackendDiModuleTest.kt#L163-L174): this test resolves `AgentToolCatalog` with the default OAuth-enabled test config, which constructs the OAuth Ktor clients, but the `finally` only closes the datasource. The production shutdown path explicitly closes these clients via `BackendRuntimeResources`; the test should use the same close path or disable OAuth config for tests unrelated to OAuth.

- [P3] [`[SkillOAuthGateway.kt#L88-L92](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/skill-oauth-api/src/main/kotlin/ru/souz/skilloauth/SkillOAuthGateway.kt#L88-L92)`](https://github.com/D00mch/souz/blob/8d4dc2f0ab7aa9ee4aedca656ed13e17d22ff3d3/skill-oauth-api/src/main/kotlin/ru/souz/skilloauth/SkillOAuthGateway.kt#L88-L92): `ApiCallResponse.headers: Map<String, String>` bakes a lossy header contract into the API. Since this is the boundary that will later become a Go service contract, either drop response headers until needed or make them `Map<String, List<String>>`.

**Simplification Options**

- Replace the custom YAML-ish parsing for `oauthScopes` with a structural YAML parser. The PR has already accumulated several parser-specific fixes; this is a sign the abstraction is becoming more expensive than a real parser.

- Consider whether `ConnectOAuthProvider` needs to exist separately. `SafeApiCall` already returns `reconnectRequired + authorizationUrl`; if explicit pre-connect UX is not required, one tool is less surface area.

- Tighten `ApiCallRequest` before it becomes a service contract: use an enum for method and either a small allowlist or documented denylist for forwarded headers. That keeps the future Go boundary simpler.

Verification: ran `git diff --check origin/main...origin/pr/615`; no whitespace issues. Full test suites were not run.

PR #626, opened 2026-08-09
feat: cross-channel messaging (ListActiveChannels, SendMessageToChannel)

jilees:
Lets the agent enumerate a user's other configured channels (currently Telegram, plus any chat reached through the public Client-Souz WebSocket contract) and forward a message into one of them mid-conversation, via a new ChannelProvider abstraction so future channel types plug in without touching call sites.

- New ChannelProvider/ChannelProviderRegistry with Telegram and public-client implementations.
- New ToolCategory.CHANNEL_MESSAGING wired through classification and the backend-safe tool allowlist.
- Public Client-Souz WS event stream now also admits out-of-band message.created pushes (executionId == null) so live-connected public clients see forwarded messages immediately.


D00mch:
@codex review

D00mch:
@codex review

D00mch:
@codex review

D00mch:
## Consolidated simplification review

Reviewed head `4982f771` against current `main` at `4c381c68` (#629).

### Start with the rebase: much of this PR can disappear

The backend now builds only `AgentId.SKILLS_GRAPH` and inventories compiled tool IDs directly ([current AgentExecutionKernel.kt](https://github.com/D00mch/souz/blob/4c381c681a4b773eacf053d1d55ad1f2895683b3/agent/src/main/kotlin/ru/souz/agent/AgentExecutionKernel.kt#L79), [NodesSkillInventory.kt](https://github.com/D00mch/souz/blob/4c381c681a4b773eacf053d1d55ad1f2895683b3/agent/src/main/kotlin/ru/souz/agent/nodes/NodesSkillInventory.kt#L84)). Backend execution no longer uses `NodesClassification` or `LocalRegexClassifier`.

That makes the new repo-wide `CHANNEL_MESSAGING` category and its fan-out unnecessary: the enum entry, classifier prompt/regexes/tests, and empty branches in desktop and both shared runtime catalogs can all be removed. Put `ListActiveChannels` and `SendMessageToChannel` under the existing `CHAT` category; the skills inventory already exposes their exact IDs.

The PR also still wires abstractions deleted by #629, including `BackendMergedToolCatalog`, `BackendSkillCoreToolsFactory`, `BackendClientToolCatalogFactory`, and `ToolRunSkillCommand` ([PR BackendDiModule.kt](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/app/BackendDiModule.kt#L239)). Please do not restore them. A single backend-specific catalog can decorate `RuntimeToolsFactory` with the two `CHAT` tools and remain the canonical `AgentToolCatalog` injected into settings, bootstrap, and runtime execution.

Keep these tools on the compiled side of `BackendExecutionToolCatalog`, so Telegram/proxy turns can use them and the persisted enabled-tool snapshot still applies. Putting them into `BackendClientSkills` would make them Client-Souz-only and bypass that policy.

### Correctness findings

1. **[P1] The forwarded message never enters the target agent's context.**

   [`persistChannelMessage`](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt#L47-L61) writes a visible `ASSISTANT` message and event, but not the target `AgentSessionRepository.history`. The next turn is seeded only from persisted session history ([BackendConversationRuntime.kt](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/agent/runtime/conversation/BackendConversationRuntime.kt#L52-L55)), not from the `messages` table. A user replying in the target channel therefore talks to an agent that cannot see what was forwarded.

   If the target already has an active execution, its finalizer can then advance `basedOnMessageSeq` past the forwarded row while saving a session that omits it ([AgentExecutionFinalizer.kt](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/execution/service/AgentExecutionFinalizer.kt#L316-L324)). This behavior is unchanged on current main.

   Please decide the domain semantics explicitly: either make forwarding a notification rather than a conversation message, or route it through one target-chat operation that serializes with active execution and updates message, session, chat timestamp, and event together.

2. **[P1] Message and durable event are committed independently, and the chat timestamp is not updated.**

   The two calls at [ChannelProvider.kt:54-61](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt#L54-L61) each open their own transaction. The message can commit while event persistence fails, leaving product history and public replay/live delivery inconsistent. Concurrent sends can also assign message order and event order differently. For Telegram, the external message may already have been delivered before this failure, so retrying can duplicate it. The path also skips the normal `chats.updated_at` update, leaving the target stale in chat ordering.

   This should be one chat-domain persistence transaction, with the committed event published afterward.

3. **[P2] A disabled Telegram feature still claims public chats, and listing hides N+1 database reads.**

   The public provider's claim callback is always active ([BackendDiModule.kt:171-182](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/app/BackendDiModule.kt#L171-L182)), while the Telegram provider is registered only when the feature flag is on ([lines 348-353](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/app/BackendDiModule.kt#L348-L353)). With Telegram disabled, an existing linked binding can therefore make a valid public-client chat disappear from every provider.

   The same callback performs `getByChat` once per public chat ([PublicClientChannelProvider.kt:26-29](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/PublicClientChannelProvider.kt#L26-L29)); Telegram listing performs `getById` once per binding ([TelegramChannelProvider.kt:24-33](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/TelegramChannelProvider.kt#L24-L33)). This is hidden I/O in loops. Batch eligible public chats and active Telegram summaries, then merge in memory. That also fixes the already-open “filter after the 50-chat limit” finding.

4. **[P2] Routing remains stringly typed and the requested public type is dropped.**

   `ChannelProviderRegistry.send` selects a provider with `channelType`, then calls `sendMessage` without that type ([ChannelProvider.kt:36-38](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt#L36-L38)). The public provider only checks that the stored type belongs to the set of forwardable types ([PublicClientChannelProvider.kt:31-36](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/PublicClientChannelProvider.kt#L31-L36)), not that it equals the requested type. As soon as a second public client type is added, a `mobile_app` request plus that other type's chat ID is accepted.

   Every current destination is already identified by the same backing chat UUID. The simpler and safer tool contract is `SendMessageToChannel(channelId, text)`; the server derives the effective transport from owned state. This removes type spoofing, duplicate claim logic, and the non-canonical UUID self-send bug in one place. If provider extensibility is a committed requirement, at least make the registry own precedence/deduplication and pass a typed target through to the provider.

5. **[P2] Telegram delivery still has partial/unknown states that the result model cannot express.**

   The latest generic-error path records confirmed chunks, but cancellation after an earlier chunk succeeds rethrows without recording that prefix ([TelegramChannelProvider.kt:50-70](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/channels/TelegramChannelProvider.kt#L50-L70)). A timeout is also delivery-ambiguous. `Delivered | Failed` forces partial/unknown delivery into a human-readable failure string, encouraging a full retry that can duplicate the prefix.

   Model `PartiallyDelivered` / `DeliveryUnknown` explicitly and persist confirmed progress, or—if product requirements allow it—reject Telegram text over 4,096 characters to keep the operation much simpler.

6. **[P2] Chunking can split a UTF-16 surrogate pair.**

   [TelegramTextChunking.kt:27-33](https://github.com/D00mch/souz/blob/4982f7710f730dd4050b93d4853d478d0de41d2f/backend/src/main/kotlin/ru/souz/backend/telegram/TelegramTextChunking.kt#L27-L33) can put `hardEnd` between the two code units of an emoji. For example, 4,095 ASCII characters followed by `😀` produces isolated surrogates that are encoded as replacement characters while Souz persists the original text. Move the hard boundary back when it splits a pair and add that boundary test.

### Suggested smaller shape

- One backend `CrossChannelMessagingService` owns listing, target resolution, source-chat exclusion, transport choice, and delivery result.
- `ListActiveChannels` returns `{channelId, label}` (an optional type can remain display-only); `SendMessageToChannel` accepts only `channelId` and `text`.
- Listing uses bounded repository queries that filter before limiting and return safe projections; no token-bearing `select *` and no I/O in mapping/filter callbacks.
- Telegram wins when an active linked binding exists; otherwise an eligible public-client chat is used. The same predicate is used for list and send, including archive/source/ownership checks.
- One backend catalog exposes the two tools under `CHAT`; no global category/classifier/shared/desktop changes and no revived merged-catalog abstraction.
- Keep focused service tests plus one target-chat persistence/conversation test and one public WebSocket replay/serialization test. The registry/map/predicate-only tests can disappear.

If the provider abstraction is intentionally required for a known third transport, the smaller fallback is: remove `channelType` from the tool, have ordered registered providers return `sendIfAvailable(...): Result?`, let the registry deduplicate by chat ID and choose the first claiming provider, and batch provider listing data.

### Policy decision

Existing settings rows contain a materialized `enabledTools` set. Resolution intersects that set with supported names but never adds newly introduced tools. Existing users therefore will not receive these tools automatically. Please test and document deliberate opt-in, or add a one-time settings migration if the feature should become default-on.

The existing open findings about filtering after the 50-chat limit and comparing raw UUID strings still apply to the current implementation; the service/typed-ID redesign above resolves both. The local-classifier “через Telegram” finding becomes unnecessary once the obsolete backend classifier changes are dropped during rebase.

PR #636, opened 2026-08-13
Store settings on backend in postgres

D00mch:


D00mch:
Overall, I like the direction: PostgreSQL becomes the source of truth for mutable backend state, and backend pods become disposable. However, I would request changes before merging because the new multi-tenant and multi-pod model exposes several credential correctness issues.

## Blocking issues

### 1. GigaChat access tokens are process-global

`GigaRestChatAPI` reads and writes `GIGA_ACCESS_TOKEN` through JVM system properties.

This means a client created for user B may reuse an access token obtained using user A's API key. This behavior existed before, but it becomes a tenant-isolation issue now that provider clients are created with user-scoped credentials.

Please either:

* keep the access token inside the `GigaRestChatAPI` instance; or
* introduce a shared token manager keyed by a credential fingerprint.

There should also be a regression test with two different API keys returning two different access tokens.

### 2. Codex refresh is not coordinated across pods

`refreshMutex` protects only a single `CodexOAuthService` instance. Two pods can read the same rotating credentials and both call the upstream refresh endpoint before either performs the database CAS.

The CAS chooses which result is stored, but it happens too late to coordinate the upstream rotation. Depending on the provider's refresh-token behavior, the database winner may already contain a superseded refresh token.

The losing pod also waits for the winner for only about 200 ms before failing.

Please acquire a distributed lease or PostgreSQL advisory lock **before** calling the refresh endpoint:

1. Acquire the lock.
2. Reload the credentials while holding it.
3. Refresh only if they are still expired.
4. Store the new credentials.
5. Release the lock.

Other pods should wait and reload the winner instead of issuing another refresh.

### 3. `initialSeed` is a permanent fallback, not a one-time seed

Currently:

```kotlin
loadStoredCredentials() ?: seedInitialCredentials()
```

Deleting the database row therefore restores the credentials from the deployment environment on the next `load()`. As long as the environment seed remains configured, deleting the row does not revoke the credentials. The current test explicitly preserves this behavior.

Please choose one explicit semantic:

* **One-time bootstrap:** seed the empty database during startup/bootstrap and make `load()` read only from PostgreSQL.
* **Authoritative fallback:** rename it accordingly and document that revocation requires deleting the row and removing the deployment secret.

I think the one-time bootstrap model is simpler and less surprising.

## Important issues

### 4. `SharedProviderTransport` does not actually share a transport

`SharedProviderTransport` contains only an `id`, while `RuntimeProviderChatApiBuilder` ignores both `sharedTransport` and `executionContext`.

Each execution still creates provider API objects that construct their own Ktor `HttpClient(CIO)` instances, and those clients do not have an explicit close lifecycle. The test named `shared transport is reused` verifies only the identity of a marker object, not connection pooling or resource reuse.

I suggest choosing one of these approaches:

* inject a real shared `HttpClient` or provider gateway and pass credentials per request; or
* remove `SharedProviderTransport` until it represents an actual shared resource.

The first option would both simplify the factories and make HTTP resource ownership explicit.

### 5. Skill updates leave orphaned immutable data

When a skill is updated, its registration is switched to a new bundle hash, but the previous `skill_bundles`, `skill_bundle_files`, and validation records remain indefinitely.

Please either:

* add garbage collection for unreferenced bundles and obsolete validations; or
* simplify the schema to store only the active user bundle if cross-user deduplication and bundle history are not product requirements.

At minimum, there should be a test proving that an unreferenced bundle is eventually removed while a bundle referenced by another user remains intact.

### 6. Make the client-skill boundary explicit

This default weakens the boundary:

```kotlin
userSkillBundleProvider: SkillBundleProvider = skillBundleProvider
```

A future call site can accidentally use the composite provider when client tools are disabled. I would remove the default and require both providers explicitly.

Built-in skill IDs should also be reserved. Currently, a user skill colliding with a built-in ID can resolve to the user bundle when client tools are disabled and to the resource bundle when they are enabled.

## Simplification opportunities

* If cross-user bundle deduplication and version history are not required, collapse `skill_bundles`, `skill_bundle_files`, and `user_skill_registrations` into one active user-skill representation.
* If preserving malformed UTF-16 strings is not a product requirement, normalize input and store conversation knowledge as PostgreSQL `text` instead of implementing a custom UTF-16 `bytea` representation.
* Derive backend-safe tool names and categories directly from `BackendSafeToolCatalog`. Maintaining a category allowlist, a name allowlist, and the catalog `when` separately creates three sources of truth for one tool.
* Since `backend_mutable_credentials` is constrained to the single key `codex_oauth`, a dedicated Codex credentials table or singleton row would be clearer than the generic key/value abstraction.

I would merge after issues 1–3 are resolved. I would also address the HTTP-client lifecycle and orphaned skill data before a production rollout.

D00mch:
Close in favour of this https://github.com/D00mch/souz/pull/640

PR #641, opened 2026-08-17
feat: cross-channel messaging (Part 1, contracts and interfaces)

jilees:
Lets the agent enumerate a user's other configured channels (currently Telegram, plus any chat reached through the public Client-Souz WebSocket contract) and forward a message into one of them mid-conversation, via a new ChannelProvider abstraction so future channel types plug in without touching call sites.

New ChannelProvider/ChannelProviderRegistry **without** implementations.
New ToolCategory.CHANNEL_MESSAGING wired through classification and the backend-safe tool allowlist.


backend/src/main/kotlin/ru/souz/backend/channels/tool/ToolListActiveChannels.kt
```diff
@@ -0,0 +1,48 @@
...
+import ru.souz.tool.ReturnParameters
+import ru.souz.tool.ReturnProperty
+import ru.souz.tool.ToolSetup
+
+class ToolListActiveChannels(
+    private val registry: ChannelProviderRegistry,
+) : ToolSetup<ToolListActiveChannels.Input> {
+    class Input
+
+    data class ChannelJson(val channelType: String, val channelId: String, val label: String)
```
D00mch: Let's maybe reuse `backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt`? Looks like the same abstraction.

backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Aggregates all registered [ChannelProvider]s; adding a new channel type means binding one more
+ * provider here. [excludeChannelId] — normally the calling tool's own conversation id — keeps the
+ * "never forward a message back into its own conversation" rule in exactly one place instead of
+ * each caller re-deriving and re-checking it, so a future caller can't silently reintroduce the bug
+ * by forgetting the check.
+ */
+class ChannelProviderRegistry(private val providers: List<ChannelProvider>) {
+    suspend fun listAll(userId: String, excludeChannelId: String? = null): List<ChannelDescriptor> =
+        providers.flatMap { it.listChannels(userId) }.filterNot { it.channelId == excludeChannelId }
```
D00mch: This looks like pointless filtering. Is it even realistic for a user to send something from one Telegram chat to another? And if they ask from Telegram to send something via Telegram, and the message arrives in the same chat the request came from, that seems like the expected behavior.

I can see from the commits that this change was suggested by the agent. I think it got this wrong.

backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+)
+
+sealed interface ChannelSendResult {
+    data class Delivered(val detail: String) : ChannelSendResult
+    data class Failed(val reason: String) : ChannelSendResult
+}
+
+/** One implementation per channel type — see `backend/src/main/kotlin/ru/souz/backend/channels/` siblings. */
+interface ChannelProvider {
+    fun supports(channelType: String): Boolean
```
D00mch: Seems like it's easy to just have a `val channelType: String`

backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt
```diff
@@ -0,0 +1,48 @@
...
+    suspend fun sendMessage(userId: String, channelId: String, text: String): ChannelSendResult
+}
+
+/**
+ * Aggregates all registered [ChannelProvider]s; adding a new channel type means binding one more
+ * provider here. [excludeChannelId] — normally the calling tool's own conversation id — keeps the
+ * "never forward a message back into its own conversation" rule in exactly one place instead of
+ * each caller re-deriving and re-checking it, so a future caller can't silently reintroduce the bug
+ * by forgetting the check.
+ */
```
D00mch: We could simplify the docs
```kotlin
/**
 * Aggregates registered [ChannelProvider]s. Providers list destinations; the registry may exclude
 * the current source channel when a structured source identity is available.
 */
```

D00mch (commented):
I also made a PR with what I proposed
https://github.com/jilees/souz/pull/1

backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt
jilees: Forwarding message to another channel is separate action for agent, so when you ask to resend message to the same channel you got something like this:

> user: send this to telegram
> agent: *content to send*
> agent: camrade, content resent to telegram

it looks strange for me, so I agreed with claude to filter current channel off

D00mch: Here is the reason why I though this code is unnecessary.

Codex flagged this issue during review yesterday:

> [P2] Current-channel exclusion uses the wrong identity shape. [[ChannelProvider.kt (line 32)](https://chatgpt.com/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt:32)](/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/channels/ChannelProvider.kt:32) filters by channelId == excludeChannelId, and both tools pass meta.conversationId as that exclude id. In backend execution, conversationId is the Souz chat UUID, not a transport/channel identity: [[BackendConversationRuntime.kt (line 38)](https://chatgpt.com/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/agent/runtime/conversation/BackendConversationRuntime.kt:38)](/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/agent/runtime/conversation/BackendConversationRuntime.kt:38). Existing Telegram bindings already have both chatId: UUID and telegramChatId: Long: [[TelegramBotBinding.kt (line 7)](https://chatgpt.com/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotBinding.kt:7)](/private/tmp/souz-pr-641/backend/src/main/kotlin/ru/souz/backend/telegram/TelegramBotBinding.kt:7). That forces the future Telegram provider into a bad choice: use Souz chatId as channelId and block Client-Souz -> Telegram forwarding for the same chat, or use Telegram telegramChatId and fail to block Telegram -> same Telegram self-forwarding. I’d make the contract structured now: ChannelRef(channelType, channelId) plus current source channel/type in invocation metadata, or move self-exclusion into providers until the source identity exists.

I looked into the possible solutions, and the best one involved adding two more classes. Then I thought that the problem is not that critical in the first place.

D00mch (approved):

PR #643, opened 2026-08-18
Cross channel (part 2, implementation)

jilees:


D00mch:
**Findings**

- **P1: Cross-channel messages are not added to future agent context.**  
  [ChannelMessagePersistence.kt](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/channels/ChannelMessagePersistence.kt:28) appends a `messages` row and public event, but does not update `AgentSessionRepository`. Future turns are seeded from `persistedSession?.history.orEmpty()` in [BackendConversationRuntime.kt](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/agent/runtime/conversation/BackendConversationRuntime.kt:32), so a user replying in the destination chat may see the message in UI while the model has no memory of sending it. Add a contract test around “send cross-channel, then continue in target chat”.

- **P2: Telegram send/list accepts orphaned chat bindings.**  
  [TelegramChannelProvider.kt](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/channels/TelegramChannelProvider.kt:35) lists a channel when `chat == null`, and [line 49](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/channels/TelegramChannelProvider.kt:49) only rejects `archived == true`, not missing/unowned chats. Since `telegram_bot_bindings.chat_id` has no FK to `chats` in [V6__telegram_bot_bindings.sql](/private/tmp/souz-pr643/backend/src/main/resources/db/migration/V6__telegram_bot_bindings.sql:1), stale rows can be advertised or sent to before persistence fails or creates inconsistent rows. Require an owned, unarchived chat before listing or sending.

- **P2: `public_client` can target the current chat.**  
  [ToolListActiveChannels.kt](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/channels/tool/ToolListActiveChannels.kt:42) passes only `userId`, while [PublicClientChannelProvider.kt](/private/tmp/souz-pr643/backend/src/main/kotlin/ru/souz/backend/channels/PublicClientChannelProvider.kt:33) lists every forwardable public chat. From a mobile chat, the tool can send an out-of-band assistant message back into the same active chat/thread, even though the tool contract says “another channel”. Pass `meta.conversationId` through and filter the source chat.

**Simplification Options**

1. Put target validation and message/event/session persistence behind one `ChannelDeliveryService`. Providers then only discover targets and perform external delivery.
2. Add repository methods like `listActiveTelegramChannelsForUser()` / `getActiveTelegramChannel()` that join bindings to owned chats. This removes `listForUser + async getById + null fallback` and fixes the orphan issue.
3. Revert `ChannelProviderRegistry.listAll` to simple sequential `flatMap` unless provider parallelism is measured as useful. There are only two DB-backed providers here.
4. Trim large explanatory comments after adding tests for the invariants; several comments restate behavior that would be clearer as small focused tests.
5. Use `telegramTextChunks(text)` defaulting instead of passing `TELEGRAM_TEXT_LIMIT` at call sites.

Verification: `./gradlew :backend:test` passes, focused channel/Telegram tests pass, and `git diff --check` is clean. I did not post review comments to GitHub.

jilees:
> * **P2: `public_client` can target the current chat.**

We already discussed this when finalizing the contract and agreed that this is normal, expected behavior.

> 3. Revert `ChannelProviderRegistry.listAll` to simple sequential `flatMap` unless provider parallelism is measured as useful. There are only two DB-backed providers here.

Agents are like people — every platform has its own opinion. Claude explicitly flagged this to me during review.

jilees:
Re: P2 — Telegram send/list accepting orphaned chat bindings

Fixed, but not via the exact repository-join shape suggested (listActiveTelegramChannelsForUser() / getActiveTelegramChannel() joining telegram_bot_bindings to chats). Went with a slightly different split:

Added ChatRepository.getByIds(chatIds: List<UUID>): List<Chat> — Postgres where id = any(?), single batch query, plus an equivalent in-memory implementation.
Added ChannelDeliveryService.resolveTarget(userId, chatId) / .resolveTargets(userId, chatIds), which require an owned and unarchived Chat row to exist — used by every ChannelProvider, not just Telegram.
TelegramChannelProvider.listChannels now does one bindingRepository.listForUser(userId) call plus one resolveTargets(...) batch lookup instead of a getById per binding, and drops any binding whose chat is missing/unowned/archived. sendMessage runs the same resolveTarget check before attempting delivery.
Reason for the deviation: a real SQL join in TelegramBotBindingRepository would need to live behind that same interface's in-memory test double (MemoryTelegramBotBindingRepository), which today has no dependency on ChatRepository at all and is constructed standalone in a lot of unrelated tests. Making it join against chat state would mean threading a ChatRepository/MemoryChatRepository into its constructor and updating every call site that builds it — a much bigger, cross-cutting change for what's fundamentally the same fix. The getByIds + resolveTarget(s) approach gets the same outcomes (no N+1, orphaned/archived bindings excluded from both listing and sending) while keeping the binding repository single-purpose, and it generalizes to PublicClientChannelProvider's sendMessage too, not just Telegram.

The new = any(?) query was verified against a real Postgres instance (Testcontainers), not just compiled.

D00mch:
New findings:

**Findings**

- **P1: Cross-channel delivery can break an active target turn.**  
  [ChannelDeliveryService.kt](/private/tmp/souz-pr643-now/backend/src/main/kotlin/ru/souz/backend/channels/ChannelDeliveryService.kt:52) loads the target session and saves it with the current row version. If the target chat already has a running execution, that execution loaded the same row version earlier and later saves in [AgentExecutionFinalizer.kt](/private/tmp/souz-pr643-now/backend/src/main/kotlin/ru/souz/backend/execution/service/AgentExecutionFinalizer.kt:196). The channel save can advance `row_version`, so the active execution final save hits `AgentStateConflictException` and fails. This is especially relevant now that same-chat `public_client` forwarding is considered expected.

- **P2: Empty chats still lose forwarded-message context.**  
  [ChannelDeliveryService.kt](/private/tmp/souz-pr643-now/backend/src/main/kotlin/ru/souz/backend/channels/ChannelDeliveryService.kt:54) returns when no session exists. The test at [ChannelDeliveryServiceTest.kt](/private/tmp/souz-pr643-now/backend/src/test/kotlin/ru/souz/backend/channels/ChannelDeliveryServiceTest.kt:140) codifies this. A newly created public/Telegram-backed chat can receive a forwarded assistant message in `messages`/events, but the first later agent turn still seeds from `persistedSession?.history.orEmpty()` in [BackendConversationRuntime.kt](/private/tmp/souz-pr643-now/backend/src/main/kotlin/ru/souz/backend/agent/runtime/conversation/BackendConversationRuntime.kt:32), so the model will not see that message.
  
And the fix:
https://github.com/jilees/souz/pull/2

D00mch:
There are several more problems in the PR:

0. Cross-channel messages can arrive between session checkpoints and must not be skipped by the persisted cursor. The initial fix queried pending messages on every turn, adding database overhead for a rare case; the query is now performed only when a message-sequence gap is detected.

1. **P1: delivery can silently undo a rename/archive.**  
   [ChannelDeliveryService.kt:41](https://github.com/D00mch/souz/blob/d4ab6e1db556f7468bf6c37946a8adde64608a0a/backend/src/main/kotlin/ru/souz/backend/channels/ChannelDeliveryService.kt#L41) reads a chat and performs a full update. [PostgresChatRepository.kt:120](https://github.com/D00mch/souz/blob/d4ab6e1db556f7468bf6c37946a8adde64608a0a/backend/src/main/kotlin/ru/souz/backend/storage/postgres/PostgresChatRepository.kt#L120) rewrites title, archive state, creation time, and update time. A concurrent rename/archive can be reverted, and two deliveries can move `updatedAt` backwards.

   Add an atomic `touchUpdatedAt(userId, chatId, at)` that only executes `updated_at = greatest(updated_at, ?)`.

2. **P2: cancellation loses Telegram chunks already sent.**  
   [TelegramChannelProvider.kt:46-58](https://github.com/D00mch/souz/blob/d4ab6e1db556f7468bf6c37946a8adde64608a0a/backend/src/main/kotlin/ru/souz/backend/channels/TelegramChannelProvider.kt#L46) persists a partial prefix for ordinary failures, but immediately rethrows `CancellationException`. Telegram may contain delivered text that is absent from messages, events, and future context.


Fixing PR
https://github.com/jilees/souz/pull/3

PR #681, opened 2026-08-24
Break channel tool name coupling

D00mch:
## Summary
- add shared internal constants for backend channel tool names
- use the shared names from list/send channel tools to avoid peer name coupling
- preserve existing tool names and LLM-facing text

## Tests
- git diff --cached --check
- ./gradlew :backend:test


D00mch:
@codex review

PR #695, opened 2026-08-25
Bypass JetBrains Gradle plugin markers

D00mch:
## Summary
- map JetBrains Gradle plugin ids to their implementation modules with pluginManagement resolutionStrategy
- bypass the missing org.jetbrains.kotlin.jvm.gradle.plugin marker reported by GitLab
- apply the same Kotlin plugin mapping inside the build-logic included build

## Context
GitLab moved past the previous kotlin-dsl marker failure and now fails resolving org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:2.4.10. This change avoids that marker artifact and resolves org.jetbrains.kotlin:kotlin-gradle-plugin directly.

## Verification
- ./gradlew :build-logic:check
- ./gradlew :backend:installDist --no-daemon
- ./gradlew souzGateFast --configuration-cache --configuration-cache-problems=fail

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -10,7 +10,7 @@ pluginManagement {
 
     resolutionStrategy {
         eachPlugin {
-            val version = requested.version ?: return@eachPlugin
+            val version = target.version ?: requested.version ?: return@eachPlugin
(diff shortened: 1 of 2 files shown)


D00mch:
@codex review

D00mch:
@codex review

PR #704, opened 2026-08-27
Add a dedicated, message-only LLM provider for context-window compaction

jilees:
NodesSummarization currently reuses the same LLMChatAPI as the ongoing conversation for its "MEMORY DUMP" compaction call, so compaction latency and cost track whatever provider the user happens to be chatting through.

Adds LlmMessageApi, a minimal one-method chat interface split out of LLMChatAPI, and OpenAICompatibleMessageAPI, a message-only OpenAI-compatible client with no fixed provider defaults (base URL, model and optional reasoning effort all come from config). NodesSummarization and AgentExecutionKernelFactory take it as an optional second LLM, defaulting to the primary llmApi so behavior is unchanged unless SUMMARIZATION_LLM_* is configured on the backend. The compaction call goes through the same 429 retry/backoff as regular conversation calls (extracted from BackendExecutionLlmChatApi into a shared retryChat helper), it just isn't folded into per-user cumulativeUsage since it's billed under its own operator-provided key rather than the user's own provider quota.

### DIFF
--- a/backend/src/main/kotlin/ru/souz/backend/app/BackendAppConfig.kt
+++ b/backend/src/main/kotlin/ru/souz/backend/app/BackendAppConfig.kt
@@ -63,6 +63,13 @@ data class BackendProviderRetryPolicy(
     }
 }
 
+data class BackendSummarizationLlmConfig(
+    val apiUrl: String,
+    val apiKey: String,
+    val model: String,
+    val reasoningEffort: String? = null,
+)
+
 data class BackendServerConfig(
     val host: String,
     val port: Int,
@@ -139,6 +146,7 @@ data class BackendAppConfig(
     val telegramPollingMaxConcurrency: Int = 4,
     val skillOAuthTokenEncryptionKey: String? = null,
     val skillOAuthProviderCredentials: Map<String, SkillOAuthProviderCredentials> = emptyMap(),
+    val summarizationLlm: BackendSummarizationLlmConfig? = null,
     val llmLimits: BackendLlmLimits = BackendLlmLimits(),
     val providerRetryPolicy: BackendProviderRetryPolicy = BackendProviderRetryPolicy(),
 ) {
@@ -223,6 +231,7 @@ data class BackendAppConfig(
                         null
                     }
                 }.toMap(),
+                summarizationLlm = source.summarizationLlmConfig(),
                 llmLimits = BackendLlmLimits(
                     perUserConcurrentExecutions = source.intValue(
                         envKey = "SOUZ_BACKEND_LIMIT_PER_USER_CONCURRENT_EXECUTIONS",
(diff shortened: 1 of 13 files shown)


D00mch:
@codex review

backend/src/main/kotlin/ru/souz/backend/llm/RetryChat.kt
```diff
@@ -0,0 +1,46 @@
...
+import ru.souz.llms.LLMRequest
+import ru.souz.llms.LLMResponse
+import ru.souz.llms.LlmMessageApi
+
+internal const val HTTP_TOO_MANY_REQUESTS = 429
+
+private val RETRY_AFTER = Regex("""retry-after=(\d+)""", RegexOption.IGNORE_CASE)
+
+/** Retries [request] on HTTP 429 per [retryPolicy], honoring a `retry-after=<seconds>` hint in the error message. */
+internal suspend fun retryChat(
```
D00mch, line 15: This way, these functions are visible from all the   backend code.
So I would move them back into the BackendExecutionLlmChatApi.
If you don't want these fns in the BackendExecutionLlmChatApi, let's create a class to wrap them.

D00mch: The docs are usefull, let's keep them.

backend/src/main/kotlin/ru/souz/backend/llm/RetryChat.kt
```diff
@@ -0,0 +1,46 @@
...
+    val retryAfter = RETRY_AFTER.find(message)?.groupValues?.getOrNull(1)?.toLongOrNull()
+    if (retryAfter != null) return min(retryAfter, retryPolicy.backoffMaxMs)
+    return min(retryPolicy.backoffBaseMs * (attempt + 1), retryPolicy.backoffMaxMs)
+}
+
+/** Wraps a message-only LLM client with the same 429 retry/backoff as [BackendExecutionLlmChatApi]. */
+internal class RetryingLlmMessageApi(
+    private val delegate: LlmMessageApi,
+    private val retryPolicy: BackendProviderRetryPolicy,
+    private val delayMillis: suspend (Long) -> Unit = { delay(it) },
```
D00mch, line 42: `delay(it.milliseconds)`

D00mch (commented):
I think we can significantly reduce the amount of code and avoid introducing classes like `LlmMessageApi`, `RetryingLlmMessageApi`, and a dedicated configuration class.

It may be enough to have a single constant backed by an environment variable and one line that passes a different model into the summarization request.

We could simply pass a different model to `resolveChatModel` in `kotlin/ru/souz/llms/openai/OpenAICompatibleChatAPI.kt`.

Key point:

* Most likely, there will be a single OpenAI API key both on your side and in our GitLab setup.

backend/src/main/kotlin/ru/souz/backend/app/BackendAppConfig.kt
```diff
@@ -63,6 +63,13 @@ data class BackendProviderRetryPolicy(
     }
 }
 
+data class BackendSummarizationLlmConfig(
+    val apiUrl: String,
+    val apiKey: String,
+    val model: String,
```
D00mch, line 69: There is also a problem that we have to synchronize the context size. For example, the chat model may be able to handle bigger context than summarization model and vice versa.

D00mch:
In favour of https://github.com/D00mch/souz/pull/707

PR #723, opened 2026-09-05
Simplify client response construction and graph execution delegate

D00mch:
Client rejection responses and HTTP/WebSocket thread status duplicated construction logic, while graph execution had an interface with only one implementation. This refactoring shares that code while preserving the wire contract and both agent modes.

- Put rejection factories in each acknowledgement class's companion object and use them from the service and WebSocket routes.
- Flatten the shared HTTP status payload into WebSocket status frames with `@JsonUnwrapped`.
- Collapse `GraphExecutionDelegate` and its sole implementation into one concrete class.
- Verify status field parity, explicit nulls, and rejection correlation identifiers in WebSocket regression tests.

Removes 67 net production lines; 24 net lines overall including tests and documentation.

Validation:
- Full JVM test and coverage graph passed before the companion-factory relocation.
- Backend tests and `souzGateFast` passed after the relocation.
- Local duplication detection remains at baseline; the authoritative duplication gate runs in GitHub Actions.
- `git diff --check` passed.


D00mch:
@codex review

Issue #619, opened 2026-08-06
Long running jobs / subagents (on backend)


D00mch:
## Dependency

Blocked by #701, which defines the host-independent subagent execution model. Backend work must reuse that model rather than introduce separate subagent semantics.

## Scope

Provide durable backend hosting for long-running agent and subagent work:

- persist parent/child execution relationships and lifecycle state;
- continue independently of the initiating HTTP or WebSocket connection;
- define restart recovery or safe terminalization after process failure;
- propagate cancellation between parent and child executions;
- persist and deliver progress, results, and failures to the parent conversation;
- coordinate execution ownership across backend instances, including leases where needed;
- cover disconnect, restart, cancellation, duplicate-delivery, and ownership-isolation behavior.

## Out of scope

The subagent graph, parent suspension semantics, tool selection, skill-triggered spawning, and result model belong to #701.

PR #727, opened 2026-09-06
Add awaited subagent execution

### LINKED ISSUE #701: Subagents
Subagents will be added to Souz.

While a subagent is running, the thread (agent loop) will be suspended until the subagent finishes its work.

A subagent is a regular graph implementing `Agent`. We will need a lightweight implementation with no unnecessary nodes and a limited set of tools. The tools available to a subagent are passed through its constructor. Which tools to provide is determined either by the agent or by a skill.

It will also be possible to add skills to Souz that specify that a subagent should be spawned and given a particular set of skills.

## Scope boundary

This issue owns the host-independent subagent execution model:

- subagent lifecycle and parent/child semantics;
- suspension and resumption of the parent agent loop;
- tool and skill selection for a subagent;
- result and failure propagation back to the parent.

Backend durability, recovery, distributed coordination, and long-running-job hosting are tracked in #619. That implementation must reuse the execution model defined here rather than introduce a second backend-specific subagent model.

D00mch:
Closes #701.

Adds `SpawnSubagent` to both parent agent graphs. The parent suspends while a fresh child graph runs a self-contained task with explicitly selected enabled tools or file-backed Skills, then resumes with a structured result or error. Parent cancellation propagates to the child, and tool batches are not retried after possible side effects.

- Inherit the parent's model, temperature, and context size; allow available model overrides within the same provider. Limit child execution to 32 model turns by default, configurable from 1 to 128.
- Isolate child history and intermediate output, prevent nested spawning, and preserve host permissions, client-tool routing, invocation metadata, and usage accounting on desktop and backend.
- Load selected Skill instructions through the host approval policy and scope `RunSkillCommand` to those Skills, using existing command directories and persistent outputs.
- Add lifecycle, capability selection, cancellation, retry, Skill execution, and backend integration tests, plus usage and maintenance documentation.

Backend durability and recovery remain scoped to #619.

## Validation

- JVM test suites and Kover XML/HTML reports passed; local reports were generated after the final source changes.
- `souzDuplicationCheck` matches the reviewed baseline: 511 production duplicated tokens and 0 test duplicated tokens. Exact-checkout authority is provided by CI.
- `git diff --cached --check` passed.


D00mch:
@codex review

D00mch:
@codex review

Issue #734, opened 2026-09-09
Update Bruno collection for the multi-chat WebSocket contract


D00mch:
The Bruno bundle in `docs/public-souz-contract/bruno/` still targets the single-chat WebSocket and contains a hardcoded chat ID and derived socket URL. It cannot exercise the multi-chat contract at `/v1/ws?clientType=backend`.

Track this as a follow-up to #732. Related contract workstream: #716.

## Scope

- Update `Client-Souz Contract.yml` and `local.json` together. Add a dedicated multi-chat request and retain HTTP creation, thread status, and the single-chat compatibility request.
- Add named JSON templates for creating two chats/users, creation retries, interleaved submissions, history, client tool results, cancellation, and explicit reconnect subscriptions with independent cursors.
- Remove persisted chat/thread/tool IDs and hardcoded derived URLs from portable defaults. Keep request IDs stable for retries and explain when to change them for a new operation.
- Add a concise Bruno README covering import, selecting `local`, Docker backend setup, Health, copying returned IDs, responding before tool deadlines, and reconnect/replay. Keep provider credentials in backend configuration.

Use `docs/public-souz-contract/README.md`, `openapi.yaml`, and `examples/happy-path.jsonl` as the source of truth. Match the current subscription semantics: automatic subscriptions are live-only; explicit `afterSeq` replaces that chat's stream and replays `seq > afterSeq` with `duplicate:false`; an omitted cursor preserves an existing stream with `duplicate:true`. This supersedes the older repeated-subscribe/no-op wording in #732.

## Acceptance criteria

- [ ] A fresh import works with the bundled `local` environment; its defaults match `local.json`, and all variables used by templates are defined.
- [ ] All HTTP bodies and WebSocket templates, after variable substitution, validate against the current public schemas. Cursor values are nonnegative JSON integers.
- [ ] Two chats belonging to different users work over one Bruno WebSocket, with acknowledgements, thread IDs, tool calls, and events routed by `chatId`.
- [ ] Creation retries, including HTTP↔WebSocket retries, return the same chat. Duplicate submits retain the original receipt without starting another execution.
- [ ] Automatic subscriptions deliver live events, including events caused by the triggering submit, without replaying earlier events.
- [ ] Reconnect restores both chats with independent exclusive cursors and no input submission. Explicit replay and omitted-cursor duplicate subscriptions follow the semantics above.
- [ ] History, client tool results, cancellation, HTTP creation/status, and the legacy single-chat endpoint have usable examples.
[...]

PR #735, opened 2026-09-09
Update multi-chat WebSocket examples and Bruno workflow

D00mch:
The Bruno bundle targets a stored single-chat socket and reuses message request IDs, making repeated manual workflows return duplicate receipts. Add a multi-chat request for two users, fresh `{{$randomUUID}}` IDs for every message submission, client-tool replies, and reconnect subscriptions with independent cursors. Chat creation keeps stable IDs so HTTP and WebSocket creation reuse the same chats.

Clear persisted chat IDs and derived URLs, align the bundled and standalone environments, document the everyday workflow, and shorten the public example trace. HTTP creation/status and the single-chat compatibility request remain available.

Partially addresses #734: additional retry, history, and cancellation scenarios are outside this update.

## Validation

- Validated 14 Bruno request bodies and 42 example trace frames against the public schemas; checked environment parity, local documentation links, and `git diff --check`.
- Direct Docker/PostgreSQL smoke test passed: real model responses for both chats, a `device.media.open` round trip, fresh UUID submissions, HTTP creation/status, independent replay cursors, and single-chat compatibility.
- Bruno UI import/smoke remains unverified because the import dialog did not open through automation. `user.ask` was schema-validated only.


D00mch:
@codex review

PR #736, opened 2026-09-10
Fix Kaniko snapshots of backend build JVM files

D00mch:
Kaniko can fail after a successful `:backend:installDist` when a build JVM removes `/tmp/hsperfdata_<user>/<pid>` during the filesystem snapshot. Pass `-XX:-UsePerfData` through `JAVA_TOOL_OPTIONS` on the Dockerfile's Gradle command to prevent those temporary files, preserve existing JVM options, and propagate the setting to child JVMs. The setting is scoped to the build command, leaving the runtime image's JVM monitoring unchanged.

Document the build constraint and the need to apply it to customized build Dockerfiles.

Validation:
- Java 21 smoke check confirmed `UsePerfData=false` and preservation of an existing `JAVA_TOOL_OPTIONS` property.
- `./gradlew souzGateFast` completed successfully with an existing advisory `SuspendFunSwallowedCancellation` warning in unchanged `AgentEventService.kt:143`.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck` completed; production and test duplication totals match their baselines. Exact-checkout authority is reserved for GitHub Actions.
- Documentation links resolve; `git diff --check` passed.

The Kaniko image-build job still needs to be rerun to verify the snapshot and image push. Application runtime tests were not run locally for this Dockerfile-only behavior change.

### DIFF
--- a/backend.Dockerfile
+++ b/backend.Dockerfile
@@ -5,7 +5,9 @@ WORKDIR /src
 
 COPY . .
 
-RUN ./gradlew :backend:installDist --no-daemon
+# Keep temporary JVM perfdata files out of Kaniko's post-build snapshot.
+RUN JAVA_TOOL_OPTIONS="${JAVA_TOOL_OPTIONS:-} -XX:-UsePerfData" \
+    ./gradlew :backend:installDist --no-daemon
 
 FROM eclipse-temurin:21-jre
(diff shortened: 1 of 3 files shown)


D00mch:
@codex review

Issue #731, opened 2026-09-08
Configure available subagent models using arbitrary model IDs


D00mch:
Blocked by #701 (Subagents). Follow-up to #727.

Control which models `SpawnSubagent` exposes and accepts through host configuration, including model IDs that are not declared in `LLMModel`.

The tool input in `agent/src/main/kotlin/ru/souz/agent/SubagentTool.kt` already accepts `model: String?`, but `SubagentToolFactory` resolves overrides through `resolveChatModel` and an `availableModels: () -> List<LLMModel>` callback. This prevents using a new provider model or custom deployment name through configuration alone.

### Recommended implementation option

Use a provider-scoped allowlist of plain model strings, loaded through the existing host configuration/settings mechanism. Illustrative configuration:

```json
{
  "subagents": {
    "models": {
      "OPENAI": ["my-custom-model", "my-deployment/v2"],
      "ANTHROPIC": ["my-anthropic-model"]
    }
  }
}
```

- Inject the configured choices into the spawn-tool factory and tool schema. Keep configuration IO in the hosts; expose only resolved values to `:agent`. Prefer a small value/callback over another model registry abstraction.
- Use the parent's execution provider to select the list. Preserve the current same-provider restriction; provider identity must remain explicit rather than being inferred by resolving the model string through `LLMModel`.
- Advertise the effective choices in the `model` parameter schema and validate against that same list before starting the child. Accept any configured nonblank model ID, even when absent from `LLMModel`, and preserve its spelling in the provider request.
- When configuration is absent for a provider, retain existing behavior. An explicitly configured list is authoritative, including an empty list. Omission of `model` continues to select the parent model, but that inherited model must also satisfy an explicit allowlist. Return `subagent_model_unavailable` when the selected model is disallowed.
- Carry the selected provider and raw model ID through desktop/backend execution routing, including streaming. The backend's `BackendExecutionLlmChatApi` currently resolves ordinary chat requests through the enum too; allowing strings only in `SubagentToolFactory` is insufficient. Reuse host credentials, retries, and usage accounting.

Configuration authorizes a model choice; the provider remains responsible for whether that ID exists and is accessible.

### Acceptance criteria

- A custom model ID absent from `LLMModel` can be enabled by configuration and reaches the selected provider unchanged, without a source change or rebuild.
- The advertised choices and runtime validation agree. Disallowed explicit and inherited selections fail before a provider call.
[...]

PR #738, opened 2026-09-10
Configure subagent models and cross-provider routing

D00mch:
Implements #731 on top of #727, targeting `codex/subagents-701`.

`SpawnSubagent` can select exact model IDs from `SUBAGENT_MODELS_JSON` across host-supported providers. The parent’s actual model and provider are always available, including custom OpenAI deployment IDs; omitting `model` always inherits the parent. Without configuration, only the parent model is advertised.

- Validate configuration at host startup, reject duplicate ownership and unsupported providers, and use one snapshot for schema choices and execution. The parent’s provider wins when a configured ID collides with its model ID.
- Carry internal provider metadata through desktop and backend routing while preserving exact outbound model spelling, existing credentials, retries, usage accounting, cancellation, and child isolation. Unadvertised selections fail before Skill loading or child creation.
- Cover streaming and ordinary provider requests, custom OpenAI inheritance, metadata serialization, and Docker-backed backend child workflows. Document configuration, credentials, precedence, and errors.

Cross-provider selection and unconditional parent inheritance supersede the original issue’s same-provider restriction and inherited-model allowlist behavior. Public parent-model selection, embeddings, and dedicated summarization retain their existing contracts.

This branch also carries the existing local node/Skill-loading simplification and folder-brief example commits ahead of #727.

## Validation

- Prescribed JVM test-and-coverage graph passed: 968 tests passed, 146 skipped, no failures or errors; Kover XML and HTML reports generated.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck` completed; production duplication matches the 511-token baseline and test duplication remains zero. Exact-checkout authority is provided by CI.
- `git diff --check` passed.


D00mch:
@codex review

PR #739, opened 2026-09-10
Align Client-Souz tool history fields and remove public target

D00mch:
Completed assistant tool history sent as `content.type: "tool_call"` with `content.result` was rejected because Souz expected `tool_exchange` and `output`. Accept the agreed `tool_call`/`result` shape and omit `target` from public WebSocket `tool.call.started` events, including replay. Trusted-proxy `GET /v1/chats/{chatId}/events` responses retain `target: "client"`; removal is limited to the public WebSocket DTO.

History remains strict: clients must omit `content.toolCallId` and `content.target`; the old `tool_exchange`/`output` shape is rejected. Live tool events and `tool.result` retain `toolCallId` for correlation. Keep the internal stored-event `target` discriminator so existing durable events remain replayable. Update the public schema, examples, and contract documentation.

Validation:
- All 172 backend tests passed. Existing workflows cover consecutive text/tool history ACKs, tool-history retries, invalid-history rejection correlation, result projection into the next execution, and tool-event replay without `target`. The existing client-tool workflow also checks that the same event keeps `target` in the proxy HTTP response; this check failed before the fix.
- Full JVM test/coverage graph passed: 1,078 tests discovered, 146 skipped, no failures; XML and HTML coverage reports generated.
- `./gradlew souzGateFast` succeeded with the existing advisory cancellation warning in unchanged `AgentEventService.kt:143`.
- `./gradlew souzDuplicationCheck`: duplicated-token totals match the baseline; exact-checkout authority is CI-only.
- YAML syntax, 177 schema references, 43 JSONL examples, local documentation links, and `git diff --check` validated.


D00mch:
@codex review

D00mch:
@codex review

PR #751, opened 2026-09-11
Optimize web response header mapping

### LINKED ISSUE #748: WebHttpSupport: avoid intermediate collection in Headers.toMap()
## Context

`sharedLogic/src/commonJvmMain/kotlin/ru/souz/tool/web/internal/WebHttpSupport.kt`

Similar intermediate-collection pattern to `BackendLogging.kt` — `entries().associate { }` creates an intermediate `Set<Map.Entry>` from `entries()` before `associate` builds the final map.

## Location

- **`Headers.toMap()`** (line 244): `entries().associate { it.key to it.value }` — `entries()` allocates a set of entry pairs, then `associate` iterates it into a new map. Called per `getText()` and `downloadBinary()` web tool request.

## Suggested fix

Use `buildMap` and iterate headers directly:

```kotlin
private fun Headers.toMap(): Map<String, List<String>> =
    buildMap(names().size) {
        names().forEach { name -> put(name, getAll(name)) }
    }
```

Single map allocation, no intermediate set.

D00mch:
Closes #748

## Summary
- build the response-header map directly from header names
- avoid the intermediate entry set created by `entries().associate`
- preserve the non-null response header value contract

## Verification
- `./gradlew :sharedLogic:jvmTest --tests 'ru.souz.tool.web.*'`\n- `./gradlew souzGateFast`

### DIFF
--- a/sharedLogic/src/commonJvmMain/kotlin/ru/souz/tool/web/internal/WebHttpSupport.kt
+++ b/sharedLogic/src/commonJvmMain/kotlin/ru/souz/tool/web/internal/WebHttpSupport.kt
@@ -241,7 +241,10 @@ private fun oversizeBinaryBodyError(url: String, maxBytes: Int): BadInputExcepti
     return BadInputException("HTTP response body is larger than ${maxMegabytes}MB for $url")
 }
 
-private fun Headers.toMap(): Map<String, List<String>> = entries().associate { it.key to it.value }
+private fun Headers.toMap(): Map<String, List<String>> =
+    buildMap(names().size) {
+        names().forEach { name -> put(name, getAll(name).orEmpty()) }
+    }
 
 private fun exponentialRetryDelayMillis(retry: Int): Long {
     val factor = 1L shl retry.coerceAtLeast(0)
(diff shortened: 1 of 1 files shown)


sharedLogic/src/commonJvmMain/kotlin/ru/souz/tool/web/internal/WebHttpSupport.kt
```diff
@@ -241,7 +241,10 @@ private fun oversizeBinaryBodyError(url: String, maxBytes: Int): BadInputExcepti
     return BadInputException("HTTP response body is larger than ${maxMegabytes}MB for $url")
 }
 
-private fun Headers.toMap(): Map<String, List<String>> = entries().associate { it.key to it.value }
+private fun Headers.toMap(): Map<String, List<String>> =
+    buildMap(names().size) {
+        names().forEach { name -> put(name, getAll(name).orEmpty()) }
```
D00mch: We can get rid of lookups as well
```kotlin
private fun Headers.toMap(): Map<String, List<String>> = buildMap(this@toMap.names().size) {
    this@toMap.forEach { name, values -> put(name, values) }
}
```

D00mch:
@codex review

PR #759, opened 2026-09-13
Fix RepoWise base SHA for PR merge checkouts

D00mch:
The PR event's base SHA can differ from the checked-out merge commit's first parent, causing RepoWise to fail before analysis. Derive `SOUZ_PR_BASE_SHA` from `HEAD^1` before grafting and persist it through `GITHUB_ENV` so the graft, base checkout, comparisons, and reports all use the same base. Keep the `HEAD^2` validation against the expected PR head SHA and align the quality-gate documentation.

Validation passed:
- Actionlint 1.7.7 (ShellCheck disabled), YAML parsing, and Bash syntax checks.
- Temporary Git fixture covering a stale event base, downstream checkout/comparison/report arguments, and rejection of an unexpected PR head before grafting; RepoWise and report-rendering commands were stubbed.
- `python3 -m unittest quality.repowise_quality_gate_test`.
- `./gradlew souzGateFast :build-logic:check --console=plain`.
- Documentation local links and `git diff --check`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,6 @@ jobs:
       PYTHONDONTWRITEBYTECODE: "1"
       REPOWISE_NO_SAVE_KEY: "1"
       REPOWISE_SKIP_EDITOR_SETUP: "1"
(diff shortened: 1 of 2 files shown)


D00mch:
@codex review

PR #762, opened 2026-09-15
VK bot channel

jilees:


backend/src/main/kotlin/ru/souz/backend/vk/VkBotApi.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    private suspend fun rawRequest(
+        uri: String,
+        formParameters: Map<String, String>,
+    ): VkRawResponse =
+        rawTransport {
+            val request = HttpRequest.newBuilder()
+                .uri(URI.create(uri))
+                .timeout(Duration.ofSeconds(REQUEST_TIMEOUT_SECONDS))
+                .header("Content-Type", "application/x-www-form-urlencoded")
```
D00mch, line 136: Please apply the configured `SOUZ_WEB_USER_AGENT` as the `User-Agent` header in both `rawGet` (Long Poll) and `rawRequest` (VK API calls). This client creates its own Java HttpClient and neither request builder sets that header, so the existing environment override has no effect on VK requests.

D00mch:
Closed in favour of #764

PR #770, opened 2026-09-17
Declarative composite skill commands

jilees:
## Summary

Adds declarative composite commands to a Skill's manifest — a linear, parse-time-validated chain of tool calls, bundled scripts, and deterministic waits, interpreted by `CompositeCommandExecutor` and run as a single `RunSkillCommand(arguments={composite: "<name>", inputs: {...}})` call instead of a separately LLM-decided tool call per step.

This was built as a follow-up to a prior discussion here about the Skill Tool Bridge (`SOUZ_TOOL_BRIDGE_SOCK`): a `SpawnSubagent`-based redesign replaced the bridge for one skill (`tv-control`, in a downstream skills repo), but the resulting per-step LLM decisions made navigation noticeably slower than the bridge-based version. This is the proposed alternative: keep tool orchestration declarative and reviewable (no skill-authored code gets a live socket to the tool dispatcher — a step's `tool:` name is static, visible in the manifest, and validated at parse time), while collapsing the specific multi-step sequences that were costing the most turns.

Validated live against two real skills in that downstream repo: a TV-navigation skill (`locate`: screenshot + vision-model lookup in one call; `act_and_observe`: action + settle + re-read screen in one call) and a Yandex Smart Home OAuth skill (`get_state`/`control_device`: SafeApiCall + local parsing in one call). Both are in real production use.

**What's in this PR** (6 commits, not squashed):
1. **`feat(subagent): declarative composite skill commands + configurable reasoning effort`** — the core mechanism: `SkillManifest.commands`, `CompositeCommandSpec`/`CompositeStepSpec` (`:agent`), parse-time validation (`SkillBundleParser`), `CompositeCommandExecutor` (`:sharedLogic`), `SkillCommandExecutor.Args.composite`/`inputs`. Also adds `SpawnSubagent.reasoningEffort`, threaded through `AgentSettings` to `LLMRequest.Chat.reasoningEffort`, so a skill can request a cheaper/faster reasoning level for a mostly-mechanical child task.
2. **`fix(subagent): scope composite tool steps to the calling agent's own tools`** — a composite `tool:` step must not be able to reach a tool the calling agent (parent or a spawned child) wasn't itself granted. `AgentToolExecutor` records the caller's own `AgentSettings.tools` into a new `ToolInvocationMeta` attribute at dispatch time; composite steps check against it.
3. **`fix(subagent): keep a composite script step's stdout as text, not auto-parsed`** — a script step's stdout was being JSON-sniffed the same way a tool step's structured result is; that's wrong when a script's output (e.g. a pre-built API request body) needs to flow into a later tool argument verbatim as a string.
4. **`fix(subagent): correctly scope composite tool steps for the parent's own RunSkillCommand`** — commit 2's fix alone broke the parent's own composite calls: a regular compiled tool (e.g. an OAuth `SafeApiCall`-style tool) is never in `AgentSettings.tools` for the classic skills-graph parent — it's only reachable indirectly via `RunSkillCommand(skillId: "<compiledToolName>")`, resolved by `ToolInvokeSkill`'s own, separate catalog. Fixed by unioning both sources.
5. **`style: trim oversized doc comments added for composite skill commands`** — self-explanatory.
6. **`feat(backend): build SkillCommandExecutor per request so composite tool: steps can reach client tools`** — `SkillCommandExecutor` was a DI singleton with no catalog at all. For composite `tool:` steps to reach request-scoped client-websocket tools (e.g. a device-control tool exposed only for that conversation), the backend now builds `SkillCommandExecutor` fresh per turn against the same request-scoped catalog already used for its other Skill tools, instead of a build-time singleton.

## Test plan

- [x] `:agent:test`, `:sharedLogic:jvmTest`, `:backend:test` (non-Docker), `:desktopApp:test` all pass
- [x] New unit tests cover: composite parsing/validation, step execution (tool/script/wait), the two allowlist-scoping fixes (including a regression test per fix), script-output-stays-text
- [x] Live-validated on two real skills against a real device/API in a downstream deployment (see summary above)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


D00mch:
Reimplemented with astra high https://github.com/D00mch/souz/pull/771

D00mch:
@codex review

D00mch:
@codex review

PR #771, opened 2026-09-17
Add simplified declarative composite skill commands

D00mch:
## Summary

Run a declared sequence of tool calls, bundled scripts, and waits through one `RunSkillCommand` invocation, avoiding a separate LLM decision for every step. Also expose optional child `reasoningEffort` and forward it through agent settings to model requests.

This is a smaller alternative implementation of [Declarative composite skill commands (#770)](https://github.com/D00mch/souz/pull/770). It retains the manifest format, typed reference substitution, command discovery, script stdout as text, child capability isolation, and request-scoped backend client-tool access. It is not an exact behavioral replacement; the differences below affect compatibility.

## Implementation differences from #770

- Pass the invoking `ToolInvokeSkill` resolver's filtered tool snapshot directly to the composite runner. Child snapshots contain only explicitly selected compiled tools. This removes the dispatch-time tool-name metadata and avoids adding catalogs or request-scoped construction to `SkillCommandExecutor`.
- Reuse the existing script executor through a small composite execution extension. Share template traversal between manifest validation and execution, and share the restricted-tool names with subagent selection.
- Use compact scenario and table-driven tests for mixed execution, references, discovery, capability restrictions, failures, cancellation, deadlines, and reasoning-effort propagation.

## Behavioral differences from #770

| Area | #770 | This PR |
| --- | --- | --- |
| Timeout | Each script gets the default 60-second timeout; no overall composite deadline. | `timeoutMillis` bounds the entire chain, defaulting to 60 seconds and capped at 300 seconds. Two 40-second scripts require a larger explicit timeout. |
| Inputs | Missing inputs fail when referenced. | Every declared input is required before execution, including unused inputs. |
| Manifest validation | Unknown command/step fields are ignored; identifiers, references, and script paths are more permissive. | Unknown command/step fields are rejected. Input and step identifiers are checked, self-references and malformed references are rejected, waits must be nonnegative, and script paths must pass bundle-path validation (so `./scripts/run.py` is rejected). Tool names and script paths are static. |
| Tool resolution | The executor resolves and checks tools at each step using its catalog and dispatch metadata. | All tool names are checked before any step runs, using one filtered snapshot. An unavailable later tool prevents earlier side effects. Filter changes during the chain apply to the next invocation. |
| Script failures | Failures become exit code `1`, empty stdout, `timedOut=false`, and a message containing at most 500 characters of script stderr. | Preserve the script's exit code, stdout, timeout flag, and full stderr, with the failing step identified. |
| Cancellation | Composite `runCatching` converts cancellation into a failure result. | Propagate coroutine cancellation. |
| Reasoning effort | Advertise `minimal`, `low`, `medium`, and `high` without runtime validation. | Validate those four values before creating the child; omission uses provider defaults. |

The child command helper also accepts its selected compiled tool IDs, in addition to selected file-backed Skills. Those compiled tools are already directly available to the child; this adds an invocation path without granting additional capabilities.

Completed side effects are not rolled back. Tool content describing an error remains ordinary result data. See [the manifest example and execution contract](docs/composite-skill-commands.md).

## Validation

- JVM test-and-coverage graph: `./gradlew test :sharedLogic:allTests :sharedUI:allTests :koverXmlReport :koverHtmlReport koverLog -Psouz.coverage --no-parallel` (all test and coverage tasks passed, including Docker-backed backend tests; existing opt-in skips remain).
- `./gradlew souzGateFast` passed when run separately from coverage. Combining the gate with `-Psouz.coverage` exposes Kover configurations that the module-boundary policy rejects.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck`: local duplicate-code comparison passed with no increase. Exact-checkout authority is available only in GitHub Actions.
- `git diff --check` and changed documentation links passed.
- The two new composite production files have 100% line coverage. Downstream TV/Yandex skills have not been replayed against this implementation.


D00mch:
@codex review

D00mch:
@codex review

D00mch:
@codex review

D00mch:
@codex review

PR #772, opened 2026-09-17
fix(backend): stop Telegram/VK binding polls from blocking each other

jilees:
## Summary

This fixes two related problems in `TelegramBotPollingService`/`VkBotPollingService`'s production poll loop (`start()`):

1. **Head-of-line blocking across bindings.** Every poll tick re-launched *all* enabled bindings inside one shared `supervisorScope`, so the tick couldn't complete until every sibling's own long poll also returned. Telegram's/VK's long poll idles for up to ~30s/~25s when a binding has no new messages, so with more than one binding, every tick was gated by the slowest — often idle — one. In practice this meant a user could see 0–20+ seconds of dead time between sending a message and the bot even starting to process it (before "typing" ever appears), simply because some *other* user's binding happened to be idling out its long poll in the same shared batch. Fixed by giving each enabled binding its own independent, long-lived loop (`BindingLoopSupervisor`), so one binding's poll cycle can never block another's.

2. **The concurrency-limiting semaphore would have reintroduced the same bug at scale.** The existing `maxConcurrency` semaphore wrapped the *entire* per-binding poll-and-process call, including the idle long poll itself. Once the per-binding-loop fix above decouples bindings from each other, that semaphore becomes the next bottleneck: with more enabled bindings than `maxConcurrency`, an idle binding sitting in its ~30s long poll would still hold a permit the whole time, starving any binding beyond the limit out of a permit — i.e. the same head-of-line blocking, just moved from the shared `supervisorScope` to the semaphore. Fixed by moving the semaphore to wrap only the actual processing step (`handleUpdate`, inside `pollBinding`), not the idle long poll.

`pollEnabledOnce()` (the old batched-round implementation) is kept as-is for the existing tests that rely on its single-round, deterministic semantics; only the production `start()` path was changed.

## Test plan
- [x] `:backend:compileKotlin` / `:backend:compileTestKotlin`
- [x] `:backend:test` (with Docker/Testcontainers) — 197/197 passing, including `TelegramBotPollingServiceTest` (lease renewal/fencing) and the `BackendTelegramE2eTest`/`BackendVkE2eTest` e2e suites that exercise `start()`
- [x] Verified against production logs on a live deployment: before the fix, per-message delay before the "typing" indicator ranged 0–20+s depending on whether another enabled binding was mid long-poll; after the fix, the delay was consistently near-zero across several consecutive test messages

🤖 Generated with [Claude Code](https://claude.com/claude-code)


D00mch:
@codex review

D00mch:
@codex review

D00mch:
@codex review

D00mch:
@codex review

PR #773, opened 2026-09-17
Fix Telegram and VK binding polls blocking each other

D00mch:
Idle Telegram/VK long polls delay unrelated bindings because a shared polling round waits for every binding and holds concurrency permits during idle network waits. This PR keeps the two fixes from #772: an independent loop for each enabled binding, and a semaphore around update processing only.

Compared with #772:

| Area | #772 | This PR |
| --- | --- | --- |
| Shared scheduling | Generic `BindingLoopSupervisor<B>` with a mutex, plus `runBindingPollLoop`; each service keeps its own refresh loop. | One 47-line `pollBindings` function owns refresh, retries, and the job map. A single coroutine manages the map, so no mutex is needed. |
| Worker lifetime | Workers launch in the application scope, alongside the service polling job. | Workers are children of the polling job's supervisor scope, so cancelling that job also cancels and waits for its workers. |
| Restart during cancellation | Removes stale jobs immediately and replaces jobs whose `isActive` is false. | Retains cancelling jobs until `isCompleted`, preventing a replacement for the same binding ID from overlapping cleanup. |
| Lease checks | Checks ownership before waiting for a processing permit. | Checks ownership after acquiring the permit, before handling the update, covering lease loss while queued. |
| Test-only batching | Keeps both `pollEnabledOnce()` methods and VK's shared session map/mutex in production code. | Removes those batch methods and shared VK state. Existing deterministic tests call the actual per-binding poll; VK's small round/session helper lives in the test harness. Each production VK worker owns its session. |
| Telegram adapter | Keeps `AgentExecutionTelegramTurnExecutor`. | Replaces that forwarding class with a method reference. |
| Regression coverage | No test changes. | Adds two scheduler lifecycle/retry tests and one production-wired E2E test exercised for both channels: five bindings remain idle while a sixth completes a second poll and links. |

Both PRs use the same base commit. **Production source only:** #772 adds 87 lines and removes 10 (net +77); this PR adds 74 and removes 90 (net −16), leaving 93 fewer production lines than #772. These counts include comments and blank lines; tests and documentation are excluded.

Validation:

- All 200 backend tests passed, including existing lease, cursor, and channel workflow coverage.
- Full JVM test/coverage command passed: `./gradlew test :sharedLogic:allTests :sharedUI:allTests :koverXmlReport :koverHtmlReport koverLog -Psouz.coverage --no-parallel`.
- `./gradlew souzGateFast` passed separately from the coverage run.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck` completed; duplicate-code totals match the baseline. Exact-checkout authority remains CI-only.
- `git diff --check` passed.


D00mch:
@codex review

PR #786, opened 2026-09-22
Enable opt-in automatic memory recall for WebSocket executions

### LINKED ISSUE #784: Allow automatic memory recall for WebSocket executions via environment configuration
## Problem

Automatic memory recall is currently disabled when client tools are enabled. Consequently, WS clients must retrieve memory themselves or rely on the agent explicitly calling SearchMemory.

Using the WS API should not prevent a deployment from letting Souz automatically retrieve memory before answering.

## Desired outcome

Allow deployments to enable automatic memory recall for WS executions, preferably through a backend environment variable. Clients should not need to implement retrieval or change their WS request payloads.

## Acceptance criteria

- When enabled, WS executions automatically retrieve relevant memory before the first LLM call, even when client tools are enabled.
- When the setting is absent, existing behavior is preserved. Deployments whose clients already supply memory can keep automatic recall disabled.
- Recall runs once during initial execution setup. Tool-loop iterations, duplicate submissions, reconnects, and continuation of an active execution do not trigger additional initial recall.
- Explicit memory tools and memory capture remain available independently of automatic recall.
- Existing user/chat isolation, stale injected-memory cleanup, and graceful handling of Hindsight failures are preserved.

Document the environment setting, its default, and when to enable it. Cover enabled, disabled, and default behavior in WS tests.

D00mch:
WebSocket executions can opt into initial memory recall with `SOUZ_FEATURE_WS_AUTOMATIC_MEMORY_RECALL=true` when `HINDSIGHT_API_URL` is configured. The default remains disabled, and clients require no payload changes. Recall runs once before the first LLM call; tool loops, duplicate submissions, reconnects, and active-thread continuation do not repeat it. Explicit memory search, capture, owner/chat isolation, stale-memory cleanup, and graceful Hindsight failure handling remain independent of the setting.

Documents the environment variable and JVM property, and adds production-wired WS coverage for default, disabled, and enabled behavior plus execution lifecycle and failure scenarios.

Validation:

- Full JVM test and coverage graph: 1,035 passed, 146 skipped; all 212 backend tests passed. XML and HTML coverage reports generated.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck` completed; local duplication matched baseline. Exact-checkout authority is evaluated in CI.
- `git diff --check` passed.

Closes #784.


D00mch:
Closed in favour of https://github.com/D00mch/souz/pull/791

PR #787, opened 2026-09-22
Sanitize tool previews once and drop intermediate collections

### LINKED ISSUE #752: ToolCallPreviewer: sanitize tool previews once and avoid intermediate collections
## Context

Paths relative to `backend/src/main/kotlin/ru/souz/backend/`:

- `agent/runtime/BackendAgentRuntimeEventSink.kt`
- `agent/runtime/ToolCallPreviewer.kt`

Confirmed code-level opportunity; performance impact has not been benchmarked.

## Problem

Tool-start and tool-finish handling generate a preview node and then call a JSON-preview helper that generates the same preview again. This repeats conversion, sanitization, and truncation work.

Additionally, object and array sanitization materializes every entry into a list, then creates another list with `take()`, even though only the first eight entries are included in the preview.

## Proposed change

Generate each sanitized preview once and reuse it for persistence and event delivery, retaining the existing safe serialization behavior.

Iterate directly over the required object fields or array elements. Use the node’s size to calculate the omitted-entry count rather than collecting discarded entries.

Do not change redaction rules, parsing behavior, or preview limits. Remove the initial input-tree `deepCopy()` only after verifying ownership and confirming that the sanitized output cannot share mutable containers with the input.

## Acceptance criteria

- Each tool-start/tool-finish preview is sanitized once.
- Existing JSON output, field order, redaction, truncation markers, and error fallbacks remain unchanged.
- Preview persistence still works when public tool events are disabled.
- Regression coverage includes empty collections, collections at and above the limit, deep nesting, sensitive values, oversized previews, and input immutability.

5aboteur:
Closes #752

## Summary
- build each tool start/finish preview once and serialize that node for `tool_calls` persistence; the durable event reuses the same node
- iterate only the first object properties and array elements, deriving the omitted-entry count from the node size; `properties()` replaces the deprecated `fields()`
- drop the input-tree `deepCopy`: sanitization only reads the input and builds new containers
- remove the duplicate text branch in `sanitizeNode`
- redaction rules, parsing, limits, field order, truncation markers, and error fallbacks are unchanged

## Verification
- `ToolCallPreviewerTest` pins exact output for empty collections, entries at and above the limits, array and object depth, top-level and nested secret keys, secret text, long strings, oversized previews, all three placeholders, and input immutability; the same cases pass against the `main` previewer
- the execution audit E2E scenario runs with tool events on and off: the `tool_calls` row stays persisted and redacted, and with events on the event previews equal the stored ones
- `./gradlew :backend:test` (211 tests) and `./gradlew souzGateFast`
- `./gradlew souzDuplicationCheck`: duplicated tokens match the baseline (+0); exact-checkout authority runs in CI
- `git diff --check`


D00mch:
@codex review

D00mch:
I've took the good parf from this PR, and merged it into the https://github.com/D00mch/souz/pull/789

PR #802, opened 2026-09-23
Support cross-channel WebSocket tools and Orion voice commands

jilees:
Client WebSocket Skills can target another of the user's device channels with `channelId`, allowing HTTP and messaging-channel executions to use a connected device. Add the bundled `orion.call` Skill for music, playback, volume, timers and alarms; preserve server-side search outside Client-Souz executions.

Cross-channel calls use scoped, in-memory waiters in `ClientThreadRuntimeRegistry` without creating a target-chat execution. Validate target ownership and availability, strip the routing argument, and wait for the source execution's acceptance ACK before publishing. Each subscription has a bounded command queue that does not evict commands; reject publication if no subscription can accept it.

Result receipt competes atomically with timeout. An on-time result resumes the caller only after its ACK, even when that write passes the result deadline. Failed ACK writes and caller cancellation release pending work.

The caller and target subscription must share a backend process. Events have `seq: null` and a correlation-only `threadId`, with no durable replay or result receipt. The public schema and documentation describe this contract.

Validation:
- `./gradlew souzGateFast` — passed.
- `npm ci --prefix quality` and `./gradlew souzDuplicationCheck` — no new duplicated tokens; exact-checkout authority remains a CI check.
- Full JVM test/coverage graph — 1,048 passed, 146 skipped, no failures; Docker-backed backend tests and XML/HTML reports included.
- Regression coverage for source ACK ordering, command queue congestion, result/timeout races, delayed or failed ACKs, and cancellation. Shared scenarios cover WS/HTTP callers, ownership, target threads, replay exclusion, and search-provider selection.


D00mch:
@codex review

backend/src/main/kotlin/ru/souz/backend/client/BackendClientSkills.kt
```diff
@@ -152,6 +145,56 @@ private class ClientWebSocketSkill(
...
+        val chat = channelDeliveryService.resolveTarget(meta.userId, chatId)
+            ?.takeIf { it.clientType in supportedClientTypes }
+            ?: return errorMessage(functionCall.name, "client_context_missing", "Device channel not found.")
+        if (!eventService.hasLiveSubscriber(chat.userId, chat.id)) {
+            return errorMessage(functionCall.name, "client_context_missing", "No device is connected on that channel.")
+        }
+        val threadId = UUID.randomUUID()
+        val context = ToolCallContext(meta.userId, chatId.toString(), threadId.toString(), UUID.randomUUID().toString())
+        return registry.withChannelTool(context) { result ->
+            val deadlineAt = now().plus(timeout)
```
D00mch: **[P2] Wait for the source execution's acceptance ACK before publishing.**

The same-thread path calls `registry.awaitAcceptedInputAcks(threadId)`, but this cross-channel path publishes immediately. `PublicClientService.startThread` launches the execution before returning the submit ACK, so a target subscribed on another socket can receive and execute this command before the originating `message.submit` is acknowledged. The connection writer mutex cannot order writes across sockets. A focused check confirmed that this path publishes while the source ACK gate is still closed.

Before publication, await pending ACKs using the source execution ID from `meta.requestId`. The UUID created here is only the target correlation ID and has no source ACK state; HTTP callers without a registry entry can remain a no-op. Add a regression test that holds the source ACK gate closed and verifies that the target receives no command until it opens.

backend/src/main/kotlin/ru/souz/backend/client/BackendClientSkills.kt
```diff
@@ -152,6 +145,56 @@ private class ClientWebSocketSkill(
...
+            eventService.publishLive(
+                userId = meta.userId,
+                chatId = chatId,
+                executionId = threadId,
+                type = AgentEventType.TOOL_CALL_STARTED,
```
D00mch: **[P2] Avoid silently evicting the only copy of a device command.**

`publishLive` uses `AgentEventBus`, whose subscriber channels have a 256-item buffer with `DROP_OLDEST`. If a target subscriber is temporarily blocked and 256 more events for that chat arrive after this command, the command is discarded while the subscription remains registered. Unlike durable event notifications, this `seq:null` command cannot be recovered by the stream's database replay. I reproduced the eviction with the production event bus.

Unless best-effort delivery with an eventual timeout is explicitly intended, use a command delivery path that preserves the command or reports overload to the caller. A separate bounded queue is one option; durable replay is not required. Add a regression test that pauses consumption, publishes a command, floods the stream, and verifies delivery or explicit rejection.

backend/src/main/kotlin/ru/souz/backend/client/PublicClientService.kt
```diff
@@ -156,6 +156,12 @@ internal class PublicClientService(
             return rejectedTool(chat.id, threadId, toolCallId, "invalid_request", validationError, now)
         }
         val context = ToolCallContext(chat.userId, chat.id.toString(), threadId.toString(), toolCallId)
+        registry.channelTool(context)?.let { pending ->
+            return HandledClientFrame(
+                response = acceptedTool(chat.id, threadId, toolCallId, duplicate = false, now),
+                afterSend = { pending.complete(ClientToolOutcome(status, frame.result, frame.error)) },
+            )
```
D00mch: **[P2] Coordinate result acceptance with timeout before preparing an accepted ACK.**

Looking up the deferred does not reserve the result or stop the caller's timeout. This handler can prepare an accepted response just before the deadline, then wait for the socket writer while `invokeOnChannel` times out and `withChannelTool` removes the waiter. The socket subsequently sends `accepted`, and `afterSend` completes a detached deferred after the agent has already received `client_tool_timed_out`. I reproduced those inconsistent outcomes with the actual skill, registry, and result handler.

Atomically decide whether result receipt or timeout wins in the registry before preparing acceptance. Keep a separate ACK gate so accepting a result does not resume the caller before the write completes, and ensure failed writes/cancellation have a cleanup path. Add a regression test that receives a valid result before the deadline but delays its ACK beyond it; the ACK and caller outcome must agree.

D00mch (changes requested):
Reviewed head `817ef7d`. The overall approach is reasonable, but the source-ACK ordering and result/timeout races should be fixed before merging. Silent command eviction under congestion should also be addressed unless best-effort delivery with an eventual timeout is an intentional product requirement.

The three inline findings were independently reproduced with focused checks using the actual production classes and mocked repository dependencies. The existing registry and event-bus tests pass; the full Gradle suite was not run locally. GitHub checks were green when reviewed.

These fixes can preserve the scoped, in-memory design. The documented same-process requirement and absence of reconnect replay are not being treated as defects.

D00mch (approved):

